PR #27, opened 2018-08-13
aws resource names are written to tf vars file and then referenced in…

greg-snowflake:
… the terraform itself


IaC/base-config.tf
```diff
@@ -265,19 +239,14 @@ resource "aws_lambda_function" "snowalert-suppression-wrapper" {
 }
 
 resource "aws_lambda_function" "snowalert-suppression-runner" {
-  function_name = "snowalert-suppression-runner"
+  function_name = "${var.suppression_runner-name}"
```
soumitr-snowflake: this should probably be runner_name

PR #52, opened 2018-10-26
added cloudwatch metrics support for runners and handlers

soumitr-snowflake:
Two main caveats:
1. CloudWatch region is fixed to us-west-2 right now
2. CLOUDWATCH_METRICS env variable needs to be set to True

### DIFF
--- a/src/alert_suppressions_runner.py
+++ b/src/alert_suppressions_runner.py
@@ -100,6 +100,7 @@ def main():
     for squelch_name in load_rules(ctx, ALERT_SQUELCH_POSTFIX):
         run_suppressions(squelch_name)
     flag_remaining_alerts(ctx)
+
     if {CLOUDWATCH_METRICS}:
         log.metric('Run', 'SnowAlert', [{'Name': 'Component', 'Value': 'Alert Suppression Runner'}], 1)
 
--- a/src/alert_suppressions_runner.py
+++ b/src/alert_suppressions_runner.py
@@ -5,7 +5,7 @@ import uuid
 import datetime
 from typing import List
 
-from config import ALERTS_TABLE, RULES_SCHEMA, ALERT_SQUELCH_POSTFIX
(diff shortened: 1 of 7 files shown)


src/helpers/log.py
```diff
@@ -22,3 +23,15 @@ def error(*args):
 def fatal(*args):
     error(*args)
     sys.exit(1)
+
+def metric(metric, namespace, dimensions, value):
+    client = boto3.client('cloudwatch','us-west-2')
```
andrey-snowflake:
```suggestion
    client = boto3.client('cloudwatch', 'us-west-2')
    client.put_metric_data(
        Namespace=namespace,
        MetricData=[{
            'MetricName': metric,
            'Dimensions': dimensions,
            'Timestamp': datetime.datetime.utcnow(),
            'Value': value
        }]
    )

```

src/alert_suppressions_runner.py
```diff
@@ -100,6 +100,8 @@ def main():
     for squelch_name in load_rules(ctx, ALERT_SQUELCH_POSTFIX):
         run_suppressions(squelch_name)
     flag_remaining_alerts(ctx)
+    if {CLOUDWATCH_METRICS}:
```
andrey-snowflake: in Python, this is a non-empty set, which will always evaluate to Truthy. if you skip the `{}` it'll be what you prob intend

src/violation_queries_runner.py
```diff
@@ -57,6 +57,8 @@ def main():
     ctx = connect_and_execute("ALTER SESSION SET use_cached_result=FALSE;")
     for query_name in load_rules(ctx, VIOLATION_QUERY_POSTFIX):
         run_query(query_name)
+    if {CLOUDWATCH_METRICS}:
```
andrey-snowflake: ditto

PR #78, opened 2018-12-05
Now w metadatas

greg-snowflake:
Code is tested in dev and records metadata successfully.


src/runners/alert_queries_runner.py
```diff
@@ -157,12 +176,35 @@ def query_for_alerts(query_name: str):
...
+    metadata['RUN_END_TIME'] = str(metadata['RUN_END_TIME'])       # then convert to string for json serializing
+    metadata['RUN_DURATION'] = str(metadata['RUN_DURATION'])
+
+    statement = f'''
+        INSERT INTO {METADATA_TABLE}
+            (event_time, v) select '{metadata['RUN_START_TIME']}',
+            PARSE_JSON(column1) from values('{json.dumps(metadata)}')
+        '''
+    try:
+        log.info("Recording run metadata...")
```
andrey-snowflake: same as before — either period ending or "done." after it's done?

src/runners/violation_suppressions_runner.py
```diff
@@ -24,17 +31,44 @@ def run_suppression(squelch_name):
...
+        RUN_METADATA['QUERY_HISTORY'].append(metadata)
         log.fatal("Suppression query {squelch_name} execution failed.", e)
+        pass
 
     print(f"Suppression query {squelch_name} executed")
+    log.metadata_fill(metadata, status='success', rows=ctx.cursor().rowcount)
+    RUN_METADATA['QUERY_HISTORY'].append(metadata)
+
+
+def record_metadata(ctx, metadata):
```
andrey-snowflake: are these identical in the 4 files? can we put them in a helper?

greg-snowflake: record_metadata is slightly different between suppressions and queries, so we can't put it in a helper quite yet. When we get more robust failure logging for suppressions, that becomes a reasonable refactor.

src/runners/alert_queries_runner.py
greg-snowflake: fixed!

PR #93, opened 2018-12-21
query renaming through the UI works.

greg-snowflake:


src/samui/frontend/src/routes/Dashboard/Alerts.tsx
```diff
@@ -37,7 +38,21 @@ class Alerts extends React.PureComponent<AlertsProps> {
             !currentRule ? (
               'Alerts Dashboard'
             ) : currentRule.savedBody ? (
-              currentRule.title
+              <div>
+                <Input
+                  id="title_input"
+                  style={{width: 300}}
+                  defaultValue={currentRule.title}
+                  onChange={e => (currentRule.newTitle = e.target.value)}
```
andrey-snowflake: you shouldn't store state by mutating a prop of a PureComponent -- it might work because of how this version of redux is implemented and how we use it, but there is no guarantee that it will keep working if we update or won't act in ways that are super unexpected (e.g. user's typing being regularly discarded if we add a polling heartbeat to update the redux state asynchronously).

src/samui/frontend/src/routes/Dashboard/Violations.tsx
```diff
@@ -65,7 +66,21 @@ class Alerts extends React.PureComponent<AlertsProps> {
               !currentRule ? (
                 'Violations Dashboard'
               ) : currentRule.savedBody ? (
-                currentRule.title
+                <div>
+                  <Input
+                    id="title_input"
+                    style={{width: 300}}
+                    defaultValue={currentRule.title}
+                    onChange={e => (currentRule.newTitle = e.target.value)}
```
andrey-snowflake: ditto — pass it through via an action and a reducer into the redux state, then it will re-render. I know it's a little tedious to add this for simply editing a textbox, and personally prefer how e.g. Angular does two-way binding, but alas this is what we've gone with for now.

src/samui/frontend/src/routes/Dashboard/Alerts.tsx
greg-snowflake: This has been addressed!

src/samui/frontend/src/routes/Dashboard/Violations.tsx
greg-snowflake: This has been addressed!

PR #100, opened 2019-01-09
Fixes bug when more than 16k alerts need to be inserted

andrey-snowflake:
impact here is that only 16k are created at a time, and the rest are delayed until the next run. definitely worth fixing, but unlikely to impact SLA so perhaps can wait until next sprint.

not yet tested.


andrey-snowflake:
There's probably a larger refactor that can fix this and similar issues by creating a single `db.insert_alerts` instead of lots of local `log_alerts` functions.

PR #124, opened 2019-03-20
List accounts ingest

soumitr-snowflake:


src/ingestion/list_aws_accounts.py
```diff
@@ -0,0 +1,56 @@
...
+    return accounts_list
+
+def load_accounts_list(sf_client,accounts_list):
+    format_string = ", ".join(["(%s)"]*len(accounts_list))
+    snapshotclock = datetime.datetime.utcnow().isoformat()
+    sf_client.cursor().execute("insert into " + os.environ['LIST_ACCOUNTS_SNOWFLAKE_TABLE_IDENTIFIER'] + " (timestamp, account) select '" + snapshotclock + "', parse_json(column1) from values" + format_string, (accounts_list))
+
+def main():
+    current_time = datetime.datetime.now()
+    if current_time.hour = 1 and current_time.minute > 0 and current_time.minute <= 15:
```
greg-snowflake: This time check makes me a little nervous; I feel like it's more robust to query the table you load stuff into, check the timestamp of the most recent entry, and run if the newest entry is older than some threshold.

that said, I'm not the one using this data so if it's important that the data be captured at some specific time, ignore me.

soumitr-snowflake:
Updated to smart time check, env variables still need to be populated in task definition

soumitr-snowflake:
Env vars populated, task execution role updated
Should be good to go

PR #135, opened 2019-04-01
removed unneeded function

greg-snowflake:
ran the tests with this code commented out and didn't hit any errors.


andrey-snowflake:
looks like this is still used in the in-law unit that survived the refactoring fire — https://github.com/snowflakedb/snowalert/blob/master/src/runners/alert_queries_runner.py#L24-L76

andrey-snowflake:
fixed in offline CR'ed code

PR #136, opened 2019-04-02
SP-1025 Record errors in metadata (part 1)

andrey-snowflake:
This fixes Violation Queries. Rest should be a bit simpler, as a lot of this code will likely be either similar or shared.


src/runners/helpers/db.py
```diff
@@ -236,17 +238,11 @@ def insert_violations_query_run(query_name, ctx=None) -> Tuple[int, int]:
...
-    except Exception as e:
-        log.info(f"{query_name} run threw an exception:", e)
-        return 0, 0
+        result = next(fetch(INSERT_VIOLATIONS_WITH_ID_QUERY.format(**locals()), fix_errors=False))
+    except Exception:
+        log.info('warning: missing STRING ID column in RESULTS.VIOLATIONS')
+        result = next(fetch(INSERT_VIOLATIONS_QUERY.format(**locals()), fix_errors=False))
+
+    num_rows_inserted = result['number of rows inserted']
+    log.info(f"{query_name} created {num_rows_inserted} rows, updated 0 rows.")
```
greg-snowflake: Why are we specifying that it updated 0 rows? If this never has a chance of updating a row (which appears to be the case since the zero is hard-coded), this seems like extranous and confusing information to include.

PR #159, opened 2019-04-18
Add ingest for ec2 describe instances and iam credential reports

soumitr-snowflake:
Also renamed env vars for more common use friendliness

### DIFF
--- a/src/ingestion/ec2_describe_instances.py
+++ b/src/ingestion/ec2_describe_instances.py
@@ -1,0 +1,128 @@
+import boto3
+import base64
+import datetime
+import json
+import os
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
+from multiprocessing import Pool
+
+import snowflake.connector
+
+#boto3.set_stream_logger(name='botocore')
+
+INSTANCES_TABLE = os.environ['EC2_INSTANCE_LIST_SNOWFLAKE_TABLE_IDENTIFIER']
+AWS_ACCOUNTS_TABLE = os.environ['LIST_ACCOUNTS_SNOWFLAKE_TABLE_IDENTIFIER']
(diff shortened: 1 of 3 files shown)


src/ingestion/ec2_describe_instances.py
```diff
@@ -0,0 +1,128 @@
...
+    sf_client = get_snowflake_client()
+    current_time = datetime.datetime.now(datetime.timezone.utc)
+    last_time = sf_client.cursor().execute(f'SELECT max(timestamp) FROM {INSTANCES_TABLE}').fetchall()[0][0]
+    if last_time is None or (current_time - last_time).total_seconds() > 86400:
+        accounts_list = get_accounts_list(sf_client)
+        get_data(accounts_list)
+    else:
+        print("It's not time yet!")
+
+global_aws_client=get_global_aws_client()
```
andrey-snowflake: so unless there's a reason otherwise, you don't want python modules to do anything when you import them, and everything to happen after you call main(). here, you can do —

    if __name__ == '__main__':
        main()

and put `global_aws_client=get_global_aws_client()` into `get_aws_client`, since that's the only place it's used

src/ingestion/iam_credential_report.py
```diff
@@ -0,0 +1,151 @@
...
+    sf_client = get_snowflake_client()
+    current_time = datetime.datetime.now(datetime.timezone.utc)
+    last_time = sf_client.cursor().execute(f'SELECT max(timestamp) FROM {CREDENTIAL_REPORTS_TABLE}').fetchall()[0][0]
+    if last_time is None or (current_time - last_time).total_seconds() > 86400:
+        accounts_list = get_accounts_list(sf_client)
+        get_data(accounts_list)
+    else:
+        print("It's not time yet!")
+
+global_aws_client=get_global_aws_client()
```
andrey-snowflake: ditto — move this up into the function it's used and shield main() via `if __module__if __name__ == '__main__'` so it doesn't run on import

PR #160, opened 2019-04-19
fix edge case for iam password policy api call and update inserts to batch instead of parallel

soumitr-snowflake:


src/ingestion/ec2_describe_instances.py
```diff
@@ -107,31 +107,34 @@ def get_data_worker(account):
...
     start = datetime.datetime.now()
-    with Pool(processes=4) as pool:
-        pool.map(get_data_worker, accounts_list)
+    instance_list_list = list(filter(None,Pool(4).map(get_data_worker, accounts_list)))
+    instance_list = []
+    for sub_list in instance_list_list:
+        instance_list.extend(sub_list)
+    if len(instance_list):
+        sf_client = get_snowflake_client()
+        instance_groups = [instance_list[i:i+15000] for i in range(0, len(instance_list), 15000)]
```
andrey-snowflake: please use `utils.groups_of` for this, unless there's something broken about it, in which case let's fix it

src/ingestion/ec2_describe_instances.py
```diff
@@ -107,31 +107,34 @@ def get_data_worker(account):
...
 
 def get_data(accounts_list):
     start = datetime.datetime.now()
-    with Pool(processes=4) as pool:
-        pool.map(get_data_worker, accounts_list)
+    instance_list_list = list(filter(None,Pool(4).map(get_data_worker, accounts_list)))
+    instance_list = []
+    for sub_list in instance_list_list:
+        instance_list.extend(sub_list)
+    if len(instance_list):
```
andrey-snowflake: either `if instance_list` or `if len(instance_list) > 0` here

src/ingestion/iam_credential_report.py
```diff
@@ -123,31 +123,37 @@ def get_data_worker(account):
...
         data = [i.split(',') for i in report]
         reports = [dict(zip(data[0], i)) for i in data[1:]]
         policy['AccountId'] = account
         reports_json = [json.dumps({**report, "account_id": account}) for report in reports]
-        sf_client = get_snowflake_client()
+        return {'report': reports_json, 'policy': json.dumps(policy)}
 
+def get_data(accounts_list):
+    start = datetime.datetime.now()
+    results_list=list(filter(None,Pool(4).map(get_data_worker, accounts_list)))
```
andrey-snowflake: this looks like you're repeating the same thing as you have in the other file? can you add it to a common file and import from there?

src/ingestion/ec2_describe_instances.py
soumitr-snowflake: the broken part is the part where I don't know how to use utils.groups_of in random places

soumitr-snowflake:
code quality issues to be addressed soon

PR #164, opened 2019-04-26
Sp1089 refactored alert handler

rdobrik-snowflake:
Adding Slack handler


greg-snowflake:
This PRs my branch into master, which is not what we want; we want to merge your code into my branch, which we can then PR into 1.7.0 once it's tested and working.

PR #165, opened 2019-04-29
Sp1089 refactored alert handler

greg-snowflake:


.circleci/config.yml
```diff
@@ -26,6 +26,7 @@ jobs:
             pip install . && touch .venv/lib/python3.7/site-packages/snowflake/__init__.py
             pip install rpy2 pandas
             pip install mypy
+            pip install slackclient
```
andrey-snowflake: is there a need for this? circleci should not need this, nor the other installs above it?

src/runners/alert_processor.py
```diff
@@ -58,12 +58,13 @@ def get_correlation_id(ctx, alert):
     query = GET_CORRELATED_ALERT.format(time=time)
 
     try:
-        match = ctx.cursor().execute(query, [actor, object, action]).fetchall()
+        #  match = ctx.cursor().execute(query, [actor, object, action]).fetchall()
```
andrey-snowflake: vestigial comment

src/runners/alert_processor.py
```diff
@@ -58,12 +58,13 @@ def get_correlation_id(ctx, alert):
...
+        match = list(db.fetch(query, params=[actor, object, action]))
     except Exception as e:
         log.error("Failed unexpectedly while getting correlation matches", e)
         match = []
 
-    correlation_id = match[0][7] if len(match) > 0 and len(match[0]) > 7 else uuid.uuid4().hex
+    correlation_id = match[0]['CORRELATION_ID'] if len(match) > 0 and len(match[0]['CORRELATION_ID']) > 7 else uuid.uuid4().hex
```
andrey-snowflake: I think you mean to say

    'CORRELATION_ID' in match[0]

instead of

    len(match[0]['CORRELATION_ID']) > 7

src/runners/plugins/slack.py
```diff
@@ -0,0 +1,142 @@
...
+
+def handle(alert_shell, type='slack', recipient_email=None, channel=None, template=None, message=None):
+    alert = alert_shell['ALERT']
+    if not os.environ.get('SLACK_API_TOKEN'):
+        log.info(f"No SLACK_API_TOKEN in env, skipping handler.")
+        return None
+
+    slack_token = os.environ["SLACK_API_TOKEN"]
+
+    if len(slack_token) > 100:  # then it must be encrypted!
```
andrey-snowflake: is there a reason you are disabling and duplicating the logic in `kms.decrypt_if_encrypted`?

src/runners/plugins/slack.py
```diff
@@ -0,0 +1,142 @@
...
+    query = f"UPDATE results.alerts SET handled=%s WHERE alert:ALERT_ID='{alert_id}'"
+    print('Updating alert table:', query)
+    try:
+
+        db.execute(query, params=str(response))
+    except Exception as e:
+        log.error(e, f"Failed to update alert {alert_id} with status {response}")
+
+
+def handle(alert_shell, type='slack', recipient_email=None, channel=None, template=None, message=None):
```
andrey-snowflake: what is `alert_shell`? why not just pass in `alert`? it's the only thing you use...

src/runners/plugins/slack.py
```diff
@@ -0,0 +1,142 @@
...
+    alert = alert_shell['ALERT']
+    if not os.environ.get('SLACK_API_TOKEN'):
```
andrey-snowflake: to test whether a var is set use `in`. to test whether it's set to some non-falsy value, do —

    slack_token = os.environ.get('SLACK_API_TOKEN')
    if not slack_token:
        ...

src/runners/plugins/slack.py
```diff
@@ -0,0 +1,142 @@
...
+    )
+
+    log.debug(f'Slack response', response)
+
+    if response['ok'] is False:
+        log.error(f"Slack handler error", response['error'])
+        raise Exception('Slack handler error', response['error'])
+
+    if 'message' in response:
+        del response['message']
```
andrey-snowflake: why?

src/scripts/install.py
```diff
@@ -52,10 +52,16 @@ def read_queries(file):
 
 
 GRANT_PRIVILEGES_QUERIES = [
+    f'GRANT OWNERSHIP ON ALL VIEWS IN SCHEMA {RULES_SCHEMA} TO ROLE {ROLE}',
+    f'ALTER SCHEMA {DATA_SCHEMA} ENABLE MANAGED ACCESS',
+    f'GRANT SELECT ON FUTURE VIEWS IN SCHEMA {DATA_SCHEMA} TO ROLE {ROLE}',
+    # f'GRANT SELECT ON FUTURE FUNCTIONS IN SCHEMA {DATA_SCHEMA} TO ROLE {ROLE}',
+    f'ALTER SCHEMA {RULES_SCHEMA} ENABLE MANAGED ACCESS',
+    f'GRANT SELECT ON FUTURE VIEWS IN SCHEMA {RULES_SCHEMA} TO ROLE {ROLE}',
```
andrey-snowflake: this seems like something extra snuck in?

src/runners/plugins/slack.py
greg-snowflake: alert_shell is all of the columns from the alerts table, which includes ALERT, CORRELATION_ID, HANDLED, and so on. It's best practice to pass the full shell into the handler so that e.g. if the handler wants to use the CORRELATION_ID or some other metadata from the alert shell, it's available.

PR #166, opened 2019-04-30
Replaced Exception with return None to avoid retry

rdobrik-snowflake:


rdobrik-snowflake:
I am going to test it with my Docker image

andrey-snowflake:
any luck?

rdobrik-snowflake:
yes, it works fine. tested it on AWS sandbox and also with my local image agains real Salesforce data.

PR #171, opened 2019-05-04
Release v1.7.0: Handlers, Baselines, Security Hardening

andrey-snowflake:
# Custom Handlers

The Alert Handler is replaced with the Alert Dispatcher with plug-in handlers in the `./src/runners/handlers` directory, and the Jira Handler is now complemented by the Slack Handler. How to develop new ones is documented in the refreshed [CONTRIBUTING](https://github.com/snowflakedb/SnowAlert/blob/v1.7.0/CONTRIBUTING.md#developing-a-new-alert-handler) document.

# New Ingestion Scripts

We have added three new ingestion scripts —

  - `ec2_describe_instances.py`
  - `iam_credential_report.py`
  - `list_aws_accounts.py`

# Installer Improvements

The installer has several minor bug fixes, as well as new parameters which let you customize what it does —

    --admin-role to set a SnowAlert administrator other than ACCOUNTADMIN
    --nosamples to install SnowAlert without sample rules
    --config-account to auth with a named account inside your `~/.snowsql/config`

# Optimized Alert Query Runner (AQR)

We've gone through a performance review and have begun optimizing how long alerts execute to best utilize your Snowflake resources, we've also added multithreading to the AQR. Because the queries still all `MERGE` into the same table, the bulk of the benefit from this optimization is expected in a future release.

# Easier Contribution

We've rewritten the [CONTRIBUTING](https://github.com/snowflakedb/SnowAlert/blob/v1.7.0/CONTRIBUTING.md) intro and have created a more thorough automated test suite to make it easier to test changes and improvements.

# (beta) Baselines

We've added an R installation to our main container and a Baseline Runner that lets you define tables you'd like populate with statistical baselines that would take too long to include in Alert Queries on-the-fly. We've yet to sand down the corners on this feature, document the functionality, and add it to the WebUI, but we are using it internally so feel free to read the code or reach out at snowalert@snowflake.com if you'd like to help us test or develop it.

# (beta) WebUI moving towards launch

While the WebUI is still heavily a work-in-progress, it went through user feedback and an initial security hardening. The WebUI and the runners have been patched for a minor SQL injection vulnerability and other fixes that will prepare us to launch in either v1.8 or v1.9.


src/runners/alert_dispatcher.py
```diff
@@ -0,0 +1,140 @@
...
+import json
+import datetime
+import uuid
+
+from .config import CLOUDWATCH_METRICS
+from .helpers import db, log
+from .utils import apply_some, json_dumps
+
+
+def log_alerts(ctx, alerts):
```
greg-snowflake: This function never gets called; it's a leftover from when the alert_dispatcher was the alert_handler and created tickets itself; it's dead code now and can be removed.

src/runners/alert_dispatcher.py
```diff
@@ -0,0 +1,140 @@
...
+                alerts
+            )
+        except Exception as e:
+            log.error("Failed to log alert", e)
+
+    else:
+        print("No alerts to log.")
+
+
+def log_failure(ctx, alert, e):
```
greg-snowflake: Same with this function; this is vestigial code that's dead and can be removed.

src/runners/helpers/db.py
```diff
@@ -166,17 +192,31 @@ def sql_value_placeholders(n):
     return ", ".join(["(%s)"] * n)
 
 
+def insert(table, values, ovewrite=False):
```
greg-snowflake: We should fix the typoed param name here so that we don't need to reproduce the typo every time we want to overwrite data in a table.

andrey-snowflake: looks fixed

PR #173, opened 2019-05-07
Add ZenGRC Ingestion Script

greg-snowflake:
To test: Add the environment variables defined at the top of the file, then run `python zengrc.py`.

Code was tested in our DEV environment and confirmed that data ingests when scheduled, and file exits cleanly when ingestion is not scheduled.

### DIFF
--- a/src/ingestion/zengrc.py
+++ b/src/ingestion/zengrc.py
@@ -54,7 +54,7 @@ def main():
 
     # 1435 is five minutes short of one day, to prevent slow drift of ingest time.
     last_time = list(db.fetch(f"""SELECT raw FROM {ZENGRC_TABLE}
-                where EVENT_TIME < DATEADD(MINUTE, 1435, CURRENT_TIMESTAMP())"""))
+                where EVENT_TIME > DATEADD(MINUTE, -1435, CURRENT_TIMESTAMP())"""))
 
     if len(last_time) == 0:
         for e in endpoints:
--- a/src/ingestion/zengrc.py
+++ b/src/ingestion/zengrc.py
@@ -23,7 +23,7 @@ def loop(endpoint):
 
         if r.status_code != 200:
             log.error(f"Ingest request for {endpoint} failed", r.text)
-            db.ingest_request_failed(ZENGRC_TABLE, r, TIMESTAMP)
+            db.record_failed_ingestion(ZENGRC_TABLE, r, TIMESTAMP)
             break
 
         j = json.loads(r.text)
--- a/src/ingestion/zengrc.py
+++ b/src/ingestion/zengrc.py
@@ -1,5 +1,5 @@
 import requests
-import base64
+from requests.auth import HTTPBasicAuth
 import datetime
 import json
 import os
@@ -13,19 +13,13 @@ ZENGRC_TABLE = os.environ.get('ZENGRC_TABLE')
 TIMESTAMP = str(datetime.datetime.utcnow())
 
 
-def gen_headers(id, secret):
-    auth = base64.b64encode(f'{id}:{secret}'.encode()).decode()
-    return {"Authorization": f"Basic {auth}"}
-
-
 def loop(endpoint):
     log.info(f"starting {endpoint}")
     j = {'links': {'next': {'href': endpoint}}}
-    headers = gen_headers(ZENGRC_ID, ZENGRC_SECRET)
     page = 1
     while j['links']['next'] is not None:
         log.info(f"Getting page {str(page)}")
-        r = requests.get(ZENGRC_URL+j['links']['next']['href'], headers=headers)
+        r = requests.get(ZENGRC_URL+j['links']['next']['href'], auth=HTTPBasicAuth(ZENGRC_ID, ZENGRC_SECRET))
 
         if r.status_code != 200:
             log.error(f"Ingest request for {endpoint} failed", r.text)
--- a/src/ingestion/zengrc.py
+++ b/src/ingestion/zengrc.py
@@ -13,21 +13,26 @@ ZENGRC_TABLE = os.environ.get('ZENGRC_TABLE')
 TIMESTAMP = str(datetime.datetime.utcnow())
 
 
-def loop(endpoint):
+def get_next(json_body):
+    url_tail = json_body['links']['next']['href']
+    return requests.get(ZENGRC_URL+url_tail, auth=HTTPBasicAuth(ZENGRC_ID, ZENGRC_SECRET))
+
+
+def process_endpoint(endpoint):
     log.info(f"starting {endpoint}")
-    j = {'links': {'next': {'href': endpoint}}}
+    json_body = {'links': {'next': {'href': endpoint}}}
     page = 1
-    while j['links']['next'] is not None:
+    while json_body['links']['next'] is not None:
         log.info(f"Getting page {str(page)}")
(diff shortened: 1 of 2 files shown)


src/ingestion/zengrc.py
```diff
@@ -0,0 +1,67 @@
...
+from runners.helpers import db, log
+
+ZENGRC_ID = os.environ.get('ZENGRC_ID')
+ZENGRC_SECRET = os.environ.get('ZENGRC_SECRET')
+ZENGRC_URL = os.environ.get('ZENGRC_URL')
+ZENGRC_TABLE = os.environ.get('ZENGRC_TABLE')
+TIMESTAMP = str(datetime.datetime.utcnow())
+
+
+def gen_headers(id, secret):
```
andrey-snowflake: please use the auth param like we do in auth.py — https://github.com/snowflakedb/SnowAlert/blob/master/src/runners/helpers/auth.py#L42

src/ingestion/zengrc.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def loop(endpoint):
+    log.info(f"starting {endpoint}")
+    j = {'links': {'next': {'href': endpoint}}}
+    headers = gen_headers(ZENGRC_ID, ZENGRC_SECRET)
+    page = 1
+    while j['links']['next'] is not None:
+        log.info(f"Getting page {str(page)}")
+        r = requests.get(ZENGRC_URL+j['links']['next']['href'], headers=headers)
```
andrey-snowflake: please abstract out into a separate `get(path)` method that prepends ZENGRC_URL and sets auth headers

src/ingestion/zengrc.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def gen_headers(id, secret):
+    auth = base64.b64encode(f'{id}:{secret}'.encode()).decode()
+    return {"Authorization": f"Basic {auth}"}
+
+
+def loop(endpoint):
+    log.info(f"starting {endpoint}")
+    j = {'links': {'next': {'href': endpoint}}}
```
andrey-snowflake: what is `j`? is there a better name for it?

src/ingestion/zengrc.py
```diff
@@ -0,0 +1,67 @@
...
+ZENGRC_TABLE = os.environ.get('ZENGRC_TABLE')
+TIMESTAMP = str(datetime.datetime.utcnow())
+
+
+def gen_headers(id, secret):
+    auth = base64.b64encode(f'{id}:{secret}'.encode()).decode()
+    return {"Authorization": f"Basic {auth}"}
+
+
+def loop(endpoint):
```
andrey-snowflake: I would expect a function called `loop` to be really general, but here it seems relatively specific to the fact that you're using it to implement looping through pagination?. maybe `update_data` could work? in general functions should be named after the side-effects they have, where looping is an implementation detail.

src/ingestion/zengrc.py
```diff
@@ -0,0 +1,67 @@
...
+        if r.status_code != 200:
+            log.error(f"Ingest request for {endpoint} failed", r.text)
+            db.ingest_request_failed(ZENGRC_TABLE, r, TIMESTAMP)
+            break
+
+        j = json.loads(r.text)
+        data = [[json.dumps(i), TIMESTAMP] for i in j['data']]
+        if len(data) > 0:
+            try:
+                query = f"INSERT INTO {ZENGRC_TABLE} select parse_json(column1), column2 from values "
```
andrey-snowflake: pushed a minor cleanup here -- can you give me keys to test it, or test?

src/ingestion/zengrc.py
```diff
@@ -0,0 +1,64 @@
...
+
+        if r.status_code != 200:
+            log.error(f"Ingest request for {endpoint} failed", r.text)
+            db.ingest_request_failed(ZENGRC_TABLE, r, TIMESTAMP)
+            break
+
+        j = json.loads(r.text)
+        data = [[json.dumps(i), TIMESTAMP] for i in j['data']]
+        try:
+            db.insert(ZENGRC_TABLE, data, select='PARSE_JSON(column1), column2')
```
andrey-snowflake: changed this -- can you give me keys to test or test yourself?

src/runners/helpers/db.py
```diff
@@ -344,3 +349,13 @@ def record_metadata(metadata, table, e=None):
 
     except Exception as e:
         log.error(f"{record_type} metadata failed to log.", e)
+
+
+def ingest_request_failed(table, r, timestamp):
```
andrey-snowflake: functions should be called after what they do, and events (in pub/sub model) after what it is that happened. how about `record_failed_ingestion` here?

src/runners/helpers/db.py
```diff
@@ -344,3 +349,13 @@ def record_metadata(metadata, table, e=None):
...
+    log = json.dumps({'headers': dict(r.headers),
+                      'text': r.text,
+                      'status_code': r.status_code,
+                      'failure': True})
+    data = [(log, timestamp)]
+    query = f"INSERT INTO {table} SELECT PARSE_JSON(COLUMN1), COLUMN2 FROM VALUES (%s)"
```
andrey-snowflake: reuse db.insert?

src/ingestion/zengrc.py
```diff
@@ -0,0 +1,67 @@
...
+
+def main():
+    print("starting")
+    endpoints = ['/api/v2/assessments',
+                 '/api/v2/audits',
+                 '/api/v2/issues',
+                 '/api/v2/requests',
+                 '/api/v2/risks']
+
+    # 1435 is five minutes short of one day, to prevent slow drift of ingest time.
```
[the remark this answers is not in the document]
greg-snowflake: We check every fifteen minutes to see if the data is over some threshold of stale. You can imagine a scenario in which when that check happens, the data is a few seconds shy of being stale, and so ingest is delayed for fifteen minutes. 24 hours later, we check and data is a few seconds shy of being stale, so ingest is delayed for fifteen minutes (which is now 30 minutes past when it was ingested two days ago).

src/runners/helpers/db.py
greg-snowflake: db.insert doesn't generate the correct sql.

src/ingestion/zengrc.py
greg-snowflake: done

src/runners/helpers/db.py
greg-snowflake: done

src/ingestion/zengrc.py
greg-snowflake: done

src/ingestion/zengrc.py
greg-snowflake: resolved

src/ingestion/zengrc.py
greg-snowflake: resolved

src/ingestion/zengrc.py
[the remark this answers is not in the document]
andrey-snowflake: ah, got it. can update to make more clear

src/runners/helpers/db.py
andrey-snowflake: I mean, tests pass ;)

PR #180, opened 2019-05-21
adds the reserved instance ingestion scripts, which are minor modific…

greg-snowflake:
…ations on the describe instances script

### DIFF
--- a/src/ingestion/ec2_describe_reserved_instance_modifications.py
+++ b/src/ingestion/ec2_describe_reserved_instance_modifications.py
@@ -1,0 +1,154 @@
+import boto3
+import base64
+import datetime
+import json
+import os
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import serialization
+from multiprocessing import Pool
+from runners.utils import groups_of
+from runners.helpers import log
+
+import snowflake.connector
+
+INSTANCES_TABLE = os.environ['EC2_RESERVED_INSTANCE_MODIFICATIONS_LIST_SNOWFLAKE_TABLE_IDENTIFIER']
+AWS_ACCOUNTS_TABLE = os.environ['LIST_ACCOUNTS_SNOWFLAKE_TABLE_IDENTIFIER']
+AWS_AUDIT_ROLE_NAME = os.environ['AWS_AUDIT_DESTINATION_ROLE_NAME']
+SA_ACCOUNT = os.environ['INGEST_SNOWFLAKE_ACCOUNT']
+SA_USER = os.environ['INGEST_SNOWFLAKE_USER']
+SA_USER_PK = os.environ['INGEST_SNOWFLAKE_USER_PRIVATE_KEY']
+SA_USER_PK_PASSWORD = os.environ['INGEST_SNOWFLAKE_USER_PRIVATE_KEY_PASSWORD']
+CACHED_AWS_CLIENT = None
+
+def get_snowflake_client():
+    kms = boto3.client('kms')
+    password = kms.decrypt(CiphertextBlob=base64.b64decode(SA_USER_PK_PASSWORD))['Plaintext'].decode()
+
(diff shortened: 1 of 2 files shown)


src/ingestion/ec2_describe_reserved_instance_modifications.py
```diff
@@ -0,0 +1,154 @@
...
+INSTANCES_TABLE = os.environ['EC2_RESERVED_INSTANCE_MODIFICATIONS_LIST_SNOWFLAKE_TABLE_IDENTIFIER']
+AWS_ACCOUNTS_TABLE = os.environ['LIST_ACCOUNTS_SNOWFLAKE_TABLE_IDENTIFIER']
+AWS_AUDIT_ROLE_NAME = os.environ['AWS_AUDIT_DESTINATION_ROLE_NAME']
+SA_ACCOUNT = os.environ['INGEST_SNOWFLAKE_ACCOUNT']
+SA_USER = os.environ['INGEST_SNOWFLAKE_USER']
+SA_USER_PK = os.environ['INGEST_SNOWFLAKE_USER_PRIVATE_KEY']
+SA_USER_PK_PASSWORD = os.environ['INGEST_SNOWFLAKE_USER_PRIVATE_KEY_PASSWORD']
+CACHED_AWS_CLIENT = None
+
+def get_snowflake_client():
```
andrey-snowflake: why are we rewriting this instead of using db.py?

src/ingestion/ec2_describe_reserved_instance_modifications.py
```diff
@@ -0,0 +1,154 @@
...
+def get_data(accounts_list):
+    start = datetime.datetime.now()
+    instance_list_list = Pool(4).map(get_data_worker, accounts_list)
+    instance_list = [x for l in instance_list_list if l for x in l]
+    if instance_list:
+        sf_client = get_snowflake_client()
+        instance_groups = groups_of(15000,instance_list)
+        snapshot_time = datetime.datetime.utcnow().isoformat()
+        for group in instance_groups:
+            query = LOAD_INSTANCE_LIST_QUERY.format(
```
andrey-snowflake: db.py has an `insert` helper that can do this simpler, I think?

src/ingestion/ec2_describe_reserved_instance_modifications.py
```diff
@@ -0,0 +1,154 @@
...
+        log.error(f"ec2_describe_reserved_instance_modifications: account: {account[1]} exception: {e}")
+        return None
+
+def get_data(accounts_list):
+    start = datetime.datetime.now()
+    instance_list_list = Pool(4).map(get_data_worker, accounts_list)
+    instance_list = [x for l in instance_list_list if l for x in l]
+    if instance_list:
+        sf_client = get_snowflake_client()
+        instance_groups = groups_of(15000,instance_list)
```
andrey-snowflake: we're repeating this in a bunch of places -- it would be better to incorporate it into `db.insert` helper

src/ingestion/ec2_describe_reserved_instance_modifications.py
```diff
@@ -0,0 +1,154 @@
...
+                format_string=", ".join(["(%s)"] * len(group)))
+            sf_client.cursor().execute(query, group)
+    end = datetime.datetime.now()
+    log.info(f"start: {start} end: {end} total: {(end - start).total_seconds()}")
+
+def main():
+    sf_client = get_snowflake_client()
+    current_time = datetime.datetime.now(datetime.timezone.utc)
+    last_time = sf_client.cursor().execute(f'SELECT max(timestamp) FROM {INSTANCES_TABLE}').fetchall()[0][0]
+    if last_time is None or (current_time - last_time).total_seconds() > 86400:
```
andrey-snowflake: this seems it would be easier to read if we put the `current_time` into the SQL -- it's probably not going to be a bug, but every person who reads this will stop and pause to think if it might be a bug

PR #185, opened 2019-05-24
SA-k8sDeploy: Adding k8s manifests and docs

cherrera2001:
This PR adds k8s manifests for secrets, the runner, and the web ui, with an associated service.

I wrote RST documentation, but did not link it up to the index to avoid any possible merge conflicts. 
Let me know if you need any other information!

Also let me know if you are ok with the structure of `infra`


andrey-snowflake:
thank you!

PR #186, opened 2019-05-31
Sp 1133 cloudtrail connector

andrey-snowflake:


src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,211 @@
...
+    db.execute(f'DROP STAGE DATA.{name}_STAGE')
+    db.execute(f'DROP STREAM DATA.{name}_STREAM')
+    db.execute(f'DROP PIPE DATA.{name}_PIPE')
+    db.execute(f'DROP TASK DATA.{name}_TASK')
+
+    if force:
+        db.execute(f'DROP TABLE DATA.{name}_STAGING')
+        db.execute(f'DROP TABLE DATA.{name}_EVENTS_CONNECTION')
+
+    return {'deleted': 'success', 'force': force}
```
andrey-snowflake: don't return things that aren't necessary

src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,211 @@
...
+    except Exception as e:
+        return e
+
+    data = db.execute(f'DESC STAGE DATA.{name}_STAGE')
+    desc = list(data)
+    for i in desc:
+        if i[0] is 'STAGE_CREDENTIALS':
+            results.append(i)
+
+    return
```
andrey-snowflake: vestigial

src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,211 @@
...
+        db.create_table(name=name+"_EVENTS_CONNECTION", cols=CLOUDTRAIL_LANDING_TABLE, comment=comment)
+        results.append({'events_table': 'success'})
+    except Exception as e:
+        return e
+
+    try:
+        db.create_task(name=name+'_TASK', schedule='15 minutes',
+                       warehouse=WAREHOUSE, sql=cloudtrail_ingest_task)
+        results.append({'task': 'success'})
+    except Exception as e:
```
andrey-snowflake: this is really bad form (as is my auto-handling exceptions in a bunch of places) -- raising an exception means something exceptional happened and returning means everything went as expected

src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,211 @@
...
+api_version STRING,
+read_only BOOLEAN,
+resources VARIANT,
+service_event_details STRING,
+shared_event_id STRING,
+vpc_endpoint_id STRING)
+"""
+
+
+def create_connector(data):
```
andrey-snowflake: I think we're calling these `connection_options` and their type should be defined in the module, as well

PR #191, opened 2019-06-04
SP-1187 Ingestion Infrastructure

greg-snowflake:
not yet tested


src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,198 @@
...
+        results['exception'] = e
+        return results
+
+    data = db.execute(f'DESC STAGE DATA.{name}_STAGE')
+    desc = list(data)
+    for i in desc:
+        if i[0] is 'STAGE_CREDENTIALS':
+            results.append(i)
+
+    return results
```
andrey-snowflake: I know I mentioned this as a way Go does it (instead of having exceptions) but seeing it implemented makes me realize this is really un-Pythonic. let's just raise exceptions through instead of returning errors.

src/connectors/okta.py
```diff
@@ -0,0 +1,93 @@
+from runners.helpers import db, log
+from runners.config import CONNECTOR_METADATA_TABLE
+
+import datetime
+import requests
+
+CONNECTION_OPTIONS = {
+    'subdomain': {type: str},
```
andrey-snowflake: add quotes around "type" and "str"

src/connectors/okta.py
```diff
@@ -0,0 +1,93 @@
...
+    'api_key': {type: str},
```
andrey-snowflake: add `"secret": True`

src/connectors/okta.py
```diff
@@ -0,0 +1,93 @@
...
+    except Exception as e:
+        results['events_table'] = 'failure'
+        results['exception'] = e
+        return results
+
+    return results
+
+
+def populate(name, options):
+    headers = {'Accept': 'application/json', 'Content-Type': 'application/json', 'Authorization': f'SSWS {options["api_key"]}'}
```
andrey-snowflake: newlines before key-value pairs make it easier to read and also easier to diff

src/connectors/okta.py
```diff
@@ -0,0 +1,93 @@
...
+    url = f"https://{options['subdomain']}.okta.com/api/v1/logs"
+
+    timestamp_query = f"""
+    SELECT EVENT_TIME from DATA.{name}_CONNECTION
+    WHERE EVENT_TIME IS NOT NULL
+    order by EVENT_TIME desc
+    limit 1
+    """
+
+    metadata = {'START_TIME': datetime.datetime.utcnow(), 'TYPE': 'Okta', 'QUERY_NAME': f'{name}_Okta'}
```
andrey-snowflake: ditto re: newlines around k/v pairs

src/connectors/okta.py
```diff
@@ -0,0 +1,93 @@
...
+
+    try:
+        _, ts = db.connect_and_fetchall(timestamp_query)
```
andrey-snowflake: since you're discarding the ctx, simple fetch would do well here and something like

    ts = next(db.fetch(timestamp_query), [None])[0]
    if ts:
        ...

should work nicely

src/connectors/okta.py
```diff
@@ -0,0 +1,93 @@
...
+        log.info(ts)
+        ts = ts[0][0]
+        ts = ts.strftime("%Y-%m-%dT%H:%M:%S.000Z")
+        log.info(ts)
+        if len(ts) < 1:
```
andrey-snowflake: the above leaves `ts` as a string with at least two dashes so this if will always be True

src/connectors/okta.py
```diff
@@ -0,0 +1,93 @@
...
+            log.error("The okta timestamp is too short or doesn't exist; defaulting to one hour ago")
+            ts = datetime.datetime.now() - datetime.timedelta(hours=1)
+            ts = ts.strftime("%Y-%m-%dT%H:%M:%S.000Z")
+
+    except Exception as e:
+        log.error("Unable to find a timestamp of most recent okta log, defaulting to one hour ago", e)
```
andrey-snowflake: logging `e` here will log a stack trace and a traceback -- is that what we want here?

src/connectors/okta.py
```diff
@@ -0,0 +1,93 @@
...
+        ts = ts.strftime("%Y-%m-%dT%H:%M:%S.000Z")
+
+    timestamp = {'since': ts}
+    loaded = 0
+
+    while 1:
+        log.info(f"url is ${url}")
+        try:
+            r = requests.get(url=url, headers=headers, params=timestamp)
+            if str(r) != '<Response [200]>':
```
andrey-snowflake: I think you likely mean to check `r.status_code != 200`

src/runners/ingest_runner.py
```diff
@@ -15,6 +24,21 @@ def main():
...
+
+    for table in tables:
+        log.info(f"Starting {table['name']}")
+        options = yaml.load(table['comment'])
+        if 'source' in options:
+            for option in options:
+                if option.endswith('secret'):
+                    options[option] = kms.decrypt_if_encrypted(options[option])
+            connection_module = importlib.import_module(f"connectors.{options['source']}")
+            connection_module.populate(options['name'], options)
```
andrey-snowflake: `populate` makes it sound like it's wiping out the existing data. let's keep the `ingest` terminology from TDD

src/runners/ingest_runner.py
```diff
@@ -15,6 +24,21 @@ def main():
...
+        tables = db.fetch(f"SHOW TABLES LIKE '{connection_table}' IN DATA")
+    else:
+        tables = db.fetch(CONNECTION_TABLE_QUERY)
+
+    for table in tables:
+        log.info(f"Starting {table['name']}")
+        options = yaml.load(table['comment'])
+        if 'source' in options:
+            for option in options:
+                if option.endswith('secret'):
```
andrey-snowflake: use `connection_module.CONNECTION_OPTIONS[options['name']].get('secret')` here

src/runners/ingest_runner.py
```diff
@@ -15,6 +24,21 @@ def main():
...
 
+    if connection_table:
+        tables = db.fetch(f"SHOW TABLES LIKE '{connection_table}' IN DATA")
+    else:
+        tables = db.fetch(CONNECTION_TABLE_QUERY)
+
+    for table in tables:
+        log.info(f"Starting {table['name']}")
+        options = yaml.load(table['comment'])
+        if 'source' in options:
```
andrey-snowflake: let's call this "module" or "module_name" instead of "source"

andrey-snowflake: image

src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,168 @@
+from runners.helpers import db
+from runners.helpers.dbconfig import WAREHOUSE
+
+CONNECTION_OPTIONS = {
```
andrey-snowflake: so either making a manual list in `__init__.py` or having it list the directory and check the modules for `CONNECTION_OPTIONS` would help the `GET /data/connectors` endpoint along here

src/connectors/cloudtrail.py
andrey-snowflake: when successful, the return value here should be an updated version of the documentation in markdown. made [SP](https://snowflakecomputing.atlassian.net/browse/SP-1229) with details

src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,168 @@
...
+    desc = list(data)
+    for i in desc:
+        if i[0] is 'STAGE_CREDENTIALS':
+            results.append(i)
+
+    return results
+
+
+def test(name):
+    return db.fetch(f'ls @DATA.{name}_STAGE')
```
andrey-snowflake: this should be yielding things, per spec

src/runners/tests/run_alerts.py
```diff
@@ -212,7 +212,7 @@ def test_alert_runners_processor_and_dispatcher(sample_alert_rules, update_jira_
     assert len(query_rule_run_record) == 7  # 3 from samples + 4 test alert queries
 
     assert query_rule_run_record[0]['QUERY_NAME'] == 'ACTIVITY_BY_ADMIN_ALERT_QUERY'
-    assert query_rule_run_record[0]['NUM_ALERTS_CREATED'] == 47
+    assert query_rule_run_record[0]['NUM_ALERTS_CREATED'] == 48
```
andrey-snowflake: this is not the only place where this count needs to increment, although it would help a lot to make it a variable that you can bump up in just one place

src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,168 @@
...
+    data = db.execute(f'DESC STAGE DATA.{name}_STAGE')
+    desc = list(data)
+    for i in desc:
+        if i[0] is 'STAGE_CREDENTIALS':
+            results.append(i)
+
+    return results
+
+
+def test(name):
```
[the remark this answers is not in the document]
greg-snowflake: IIRC the direction was to implement `test()` fully once everything else is working, so i left what I had and didn't worry about the rest of the functionality specified in the TDD.

src/connectors/cloudtrail.py
greg-snowflake: A list of CONNECTION_OPTIONS for all of the connectors?

src/connectors/cloudtrail.py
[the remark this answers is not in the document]
andrey-snowflake: makes sense, but let's do it in parallel w/ the WebUI?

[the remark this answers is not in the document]
greg-snowflake: Sounds good, i'll work on that today.

src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,174 @@
...
+            res[i[1]] = res[i[3]]
+
+    results = []
+    for field in res:
+        data = {'title': field, 'body': res[field]}
+        results.append(data)
+
+    results.append({'title': 'Next Steps', 'body': """
+Please follow the documentation at
+https://docs.snowflake.net/manuals/user-guide/data-load-s3-config.html#step-4-configuring-the-aws-iam-role-to-allow-access-to-the-stage
```
andrey-snowflake: we agreed to rewrite this in a way that makes more sense in the context of this flow?

src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,174 @@
...
+    data = db.execute(f'DESC STAGE DATA.{name}_STAGE')
+    desc = list(data)
+    for i in desc:
+        if i[0] is 'STAGE_CREDENTIALS':
+            res[i[1]] = res[i[3]]
+
+    results = []
+    for field in res:
+        data = {'title': field, 'body': res[field]}
+        results.append(data)
```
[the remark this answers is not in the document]
greg-snowflake: It has what's specified in the TDD: the stage credentials and the link to the documentation. What would you suggest we remove?

src/connectors/okta.py
```diff
@@ -0,0 +1,101 @@
...
+    SELECT EVENT_TIME from DATA.{name}_CONNECTION
+    WHERE EVENT_TIME IS NOT NULL
+    order by EVENT_TIME desc
+    limit 1
+    """
+
+    metadata = {
+        'START_TIME': datetime.datetime.utcnow(),
+        'TYPE': 'Okta',
+        'QUERY_NAME': f'{name}_Okta'
```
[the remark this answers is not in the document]
greg-snowflake: QUERY_NAME is what db.record_metadata is looking for, so I used that in lieu of making a larger and more significant refactor of the record_metadata function.

src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,174 @@
...
+
+    pipe_sql = f"COPY INTO DATA.{name}_STAGING FROM @DATA.{name}_STAGE/"
+    db.create_pipe(name=name + '_PIPE', sql=pipe_sql, replace=True)
+
+    db.create_table(name=name + "_EVENTS_CONNECTION", cols=CLOUDTRAIL_LANDING_TABLE, comment=comment)
+
+    db.create_task(name=name + '_TASK', schedule='15 minutes',
+                   warehouse=WAREHOUSE, sql=cloudtrail_ingest_task)
+
+    data = db.execute(f'DESC STAGE DATA.{name}_STAGE')
```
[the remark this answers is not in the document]
greg-snowflake: fixed.

src/connectors/okta.py
```diff
@@ -0,0 +1,101 @@
...
+            log.error("Error with Okta logs: ", e)
+            db.record_metadata(metadata, table=CONNECTOR_METADATA_TABLE, e=e)
+
+    metadata['ROW_COUNT'] = {'INSERTED': loaded}
+    db.record_metadata(metadata, table=CONNECTOR_METADATA_TABLE)
+
+
+def process_logs(logs, name):
+    output = [(row, row['published']) for row in logs]
+    db.insert(f"DATA.{name}_CONNECTION", output, select='PARSE_JSON(column1), column2')
```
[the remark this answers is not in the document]
greg-snowflake: fixed.

src/connectors/__init__.py
```diff
@@ -0,0 +1,7 @@
+import cloudtrail
+import okta
+
+CONNECTION_OPTIONS = [
+    {'connector': 'cloudtrail', 'options': cloudtrail.CONNECTION_OPTIONS},
```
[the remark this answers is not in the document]
greg-snowflake: done

src/connectors/cloudtrail.py
```diff
@@ -0,0 +1,174 @@
+from runners.helpers import db
+from runners.helpers.dbconfig import WAREHOUSE
+
+CONNECTION_OPTIONS = {
+    'bucket_name': {'type': 'str'},
+    'filter': {'type': 'str', 'default': 'AWSLogs/'},
+    'aws_role': {'type': 'str'}
+}
```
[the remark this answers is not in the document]
greg-snowflake: fixed.

andrey-snowflake:
can you make this PR one commit from branch `SP1187_data_connectors_infra`, please?

PR #194, opened 2019-06-12
Azure connector

greg-snowflake:
Tested in oz03309 on the command line, both connect() and ingest() methods.


src/connectors/azure_ad_audit.py
```diff
@@ -0,0 +1,188 @@
...
+from azure.storage.blob import BlockBlobService
+from snowflake.ingest import SimpleIngestManager
+from snowflake.ingest import StagedFile
+
+CONNECTION_OPTIONS = [
+    {
+        'type': 'str',
+        'name': 'account_name',
+        'title': 'Storage Account',
+        'prompt': 'The storage account holding your AD Audit log blobs',
```
andrey-snowflake: default: True

greg-snowflake: fixed

src/connectors/azure_log.py
```diff
@@ -0,0 +1,333 @@
...
+    return {'newStage': 'finalized', 'newMessage': 'Table, Stage, and Pipe created'}
+
+
+def ingest(name, options):
+    block_blob_service = BlockBlobService(
+        account_name=options['storage_account'],
+        sas_token=options['sas_token'],
+        endpoint_suffix=options['suffix']
+    )
+    base_name = name.rstrip('CONNECTION')
```
andrey-snowflake: shouldn't this be `_CONNECTION`?

src/connectors/azure_log.py
```diff
@@ -0,0 +1,333 @@
...
+
+    db.execute(f"select SYSTEM$PIPE_FORCE_RESUME('DATA.{base_name}_PIPE');")
+
+    files = block_blob_service.list_blobs(options['blob_name'])
```
andrey-snowflake: can we put potential `KeyErrors` up top, so that if it errors, it does it before it tries to do anything?

src/connectors/azure_log.py
```diff
@@ -0,0 +1,333 @@
...
+                new_files.append(StagedFile(file.name, None))
+    else:
+        for file in files:
+            new_files.append(StagedFile(file.name, None))
+
+    log.info(f"Found {len(new_files)} files to ingest")
+
+    # Proxy object that abstracts the Snowpipe REST API
+    ingest_manager = SimpleIngestManager(account=environ.get('SNOWFLAKE_ACCOUNT'),
+                                         host=f'{environ.get("SNOWFLAKE_ACCOUNT")}.snowflakecomputing.com',
```
andrey-snowflake: ditto — if `environ.get('SNOWFLAKE_ACCOUNT')` is `None`, we want to error before we try anything else

src/connectors/azure_log.py
greg-snowflake: resolved.

src/connectors/azure_log.py
greg-snowflake: resolved

src/connectors/azure_log.py
greg-snowflake: resolved

src/connectors/azure_log.py
```diff
@@ -0,0 +1,374 @@
...
+        ('RESULT_TYPE', 'VARCHAR'),
+        ('TENANT_ID', 'VARCHAR'),
+        ('EVENT_TIME', 'TIMESTAMP_LTZ'),
+        ('LOADED_ON', 'TIMESTAMP_LTZ')
+    ]
+}
+
+
+def connect(name, options):
+    try:
```
andrey-snowflake: exceptions are only for exceptional behavior, and this is be definition expected, since you're catching it right away, so the right construct to use here is `if 'account_name` in options`. that said, these should already be declared in `CONNECTION_OPTIONS` as required -- can we just iterate over that, outside the connector?

src/runners/ingest_runner.py
```diff
@@ -10,14 +10,14 @@
...
 def main(connection_table="%_CONNECTION"):
-    # for name in os.listdir('../ingestion'):
-    #     log.info(f"invoking {name}")
-    #     try:
-    #         res = subprocess.call(f"python ../ingestion/{name}", shell=True)
-    #         log.info("subprocess returns: ", res)
-    #         log.info(f"{name} invoked")
-    #     except Exception as e:
-    #         log.error(f"failed to run {name}", e)
+    for name in os.listdir('../ingestion'):
```
andrey-snowflake: this change is not related to the Azure connector and duplicates one on the branch you are merging into -- could you try rebasing and squashing all these changes into a single diff?

greg-snowflake: rebased

andrey-snowflake: Are you sure? Github UI says there is still no clean merge
> This branch cannot be rebased due to conflicts
> Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

src/connectors/azure_log.py
andrey-snowflake: added these to base branch

src/runners/ingest_runner.py
greg-snowflake: hmm... Maybe? I just pushed, did that fix it?

PR #198, opened 2019-06-19
v1.8.0 Data Connectors, Orchestrator Templates, etc.

andrey-snowflake:
# Data Connectors

We have added next generation infrastructure and a UI to handle collecting data into Snowflake. Data Connectors have a standardized installation and ingestion interface.

To make sure they work best on existing installations, please add a metadata table —

    USE SCHEMA results;
    CREATE TABLE IF NOT EXISTS ingestion_metadata(event_time TIMESTAMP_LTZ, v VARIANT);
    GRANT INSERT, SELECT ON ingestion_metadata TO ROLE snowalert;

To start, we are providing three Data Connectors: Okta, CloudTrail, and Azure Cloud Logs.

# Orchestration Infrastructure

We're excited to see other companies take up SnowAlert and have preliminary data on what orchestration infrastructure our community will be using. To start, we are releasing k8s configuration files written by @cherrera2001. Thanks!

For internal teams, we've deployed SnowAlert to AWS Fargate using a Terraform module, which is included in the `infra/` directory, as well. It's minimal for what we needed so pull request are welcome for, e.g. a Scheduled Task for the ingestion runner. Thanks @rdobrik-snowflake!

# Improvements and Bug Fixes
- Fixed installer bug for custom passwords including symbol `$` (88297a8)
- Fixed installer bug on accounts w/o data sharing (49c7856)
- Fixes behavior when Alert explicitly declares empty handler list (b98a56b)
- Added `--no-samples` options to installer (30453fc)
- Updates documentation and query packs to match latest functionality (607e816, d2fdeb6)
- Fixes UDTF privilege grant in installer (a904207)
- Move Violations on/off switch to top of page and label it (2353b50)
- WebUI dependencies updated and TS lib version bumped to `es2019`


rdobrik-snowflake:
Thanks Andrey, I just updated our schema!

Roman

src/runners/helpers/vault.py
```diff
@@ -28,5 +32,15 @@ def decrypt_if_encrypted(ct: Optional[str] = None, envar: Optional[str] = None)
 
         return res['Plaintext'].decode()
```
greg-snowflake: Do we need to b64decode the plaintext here? When I use `aws kms decrypt` on the CLI, I get plaintext that's base64 encoded, and I need to decode it in order to get useful data out again.

andrey-snowflake: Yes -- I think it's the same here as it is on the CLI. In order to protect against funky stuff sneaking into the pt, KMS always b64's it. That way you're also dealing with pt that is bytes, not strings, side-stepping the need for a standard encoding:

    >>> from base64 import b64decode
    >>> b64decode('')
    b''

PR #200, opened 2019-06-21
Sa aws config data connector

cherrera2001:
This tis the initial commit of the AWS config data connector.

This connector creates three tables per connection (EC2, SG, ELB) for EC2, Security Groups, and Elastic Load Balancers respectively. 

This could be improved to also perform a VPC and EBS inventory as well. 

A few column are tabularized to ensure efficient queries for common SnowAlert rules.


src/connectors/aws_config.py
```diff
@@ -0,0 +1,348 @@
+"""AWS Config
+Collects AWS EC2 config information into a columnar table
```
greg-snowflake: This collects more than just EC2 data; it also collects security groups, elbs, and the raw snapshots.

cherrera2001: I am struggling on using this vs actual Config data snapshots...should this be a different connector?

greg-snowflake: It might be better to break these out into three individual connectors, and then also have a Config connector that ingests snapshots and creates tables for useful things that are in snapshots - that way, people can ingest precisely what they want, or they can get it all with configs.

cherrera2001: Will do!

greg-snowflake: Sorry to change this up; andrey suggested that keeping it as one connector, but allowing you to select (through a drop down) which logs you want to ingest is probably the ideal solution, similar to what's been implemented in the Azure connector. That way, we don't end up with a dozen AWS connectors in the dashboard, with one for each service. Does that seem reasonable to you?

cherrera2001: Totally, I agree with it as well. otherwise you have too much to configure at once. Will have a first pass by tomorrow.

cherrera2001: Ok after further investigation, I think it makes sense to separate this into 2 connectors, the Boto3-based asset inventory for EC2, Security Groups, and ELBS, and then another for aws config ingest. The main reason is that the configuration for the connectors are different. AWS config will setup a stage + pipe + landing table + stream + task, therefore requiring a similar configuration to cloudtrail. The Boto3-based inventory will need only an id and secret key. 

Also I would like to also make the assumption that Config will log to a single bucket for all regions in an account, otherwise there would be the need to create a lot of connectors to ingest the data. @greg-snowflake @andrey-snowflake let me know what you think.

greg-snowflake:
Shoot, I clicked the wrong button and resolved a conversation that I didn't mean to.

@cherrera2001 You've got the right idea; we'd want one connector that lets you ingest arbitrary data from AWS by using Boto (starting with the EC2 instances, ELBs, and Security Groups), and then a second connector that lets you ingest config snapshots and then create views that pull out resources from the larger table.

You're also correct that the config snapshot connector would need to create the staging table, landing table, stream, and flattening task similar to how the Cloudtrail connector sets itself up. Users can then set up multiple config connectors to ingest from different accounts if they need to, or they can collect snapshots from multiple accounts into a single bucket and just set up the connector for that bucket. 

Does that make sense? I'm happy to take on the snapshot connector if you're not quite clear on how it's supposed to work.

cherrera2001:
No problem...all sorted. I know exactly what you are after on the snapshot one.

src/connectors/aws_asset_ingest.py
```diff
@@ -0,0 +1,357 @@
...
+            row['Scheme'],
+            row['VPCId'])
+            for row in elbs],
+        select='PARSE_JSON(column1), column2, column3, column4, column5, column6, '
+               'column7, column8, column9, column10'
+    )
+    return len(elbs)
+
+
+def get_ec2_instances(client_id, secret_key, regions):
```
greg-snowflake: This doesn't record the information about which accounts had how many snapshots, but I have an implementation for that built in our current script and I can port it over to this connector far easier than I could explain it in a github comment, so this is fine for me.

src/connectors/aws_config.py
```diff
@@ -0,0 +1,348 @@
...
+    if 'SG' in table_name:
+        count = ingest_sg(client_id, secret_key, landing_table, regions)
+    if 'ELB' in table_name:
+        count = ingest_elb(client_id, secret_key, landing_table, regions)
+
+    log.info(f'Inserted {count} rows.')
+    yield count
+
+
+def ingest_ec2(client_id, secret_key, landing_table, regions):
```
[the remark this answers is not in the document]
cherrera2001: I could add that, however, we I don't have a solution that queries the organizations api yet to get the accounts info. I think it does make sense however, to record that at a higher level. I guess the question is, do we want to do this just for ec2 instances or for other types too?

src/connectors/aws_config.py
```diff
@@ -0,0 +1,248 @@
...
+    ENABLE_OCTAL = FALSE,
+    ALLOW_DUPLICATE = FALSE,
+    STRIP_OUTER_ARRAY = TRUE,
+    STRIP_NULL_VALUES = FALSE,
+    IGNORE_UTF8_ERRORS = FALSE,
+    SKIP_BYTE_ORDER_MARK = TRUE
+"""
+
+CONFIG_LANDING_TABLE_COLUMNS = [
+    ('raw', 'VARIANT'),
```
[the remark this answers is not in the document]
cherrera2001: done

src/connectors/aws_asset_ingest.py
```diff
@@ -0,0 +1,357 @@
...
+    elif options["Asset_Source"] == 'ELB':
+        result = create_config_table(connection_name, 'ELB', AWS_ELB_LANDING_TABLE_COLUMNS, options)
+
+    return {
+        'newStage': 'finalized',
+        'newMessage': f"{result}",
+    }
+
+
+def create_config_table(connection_name, config_type, columns, options):
```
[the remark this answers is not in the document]
cherrera2001: done

src/connectors/aws_asset_ingest.py
```diff
@@ -0,0 +1,357 @@
...
+    ('KEY_NAME', 'STRING(256)'),
+    ('LAUNCH_TIME', 'TIMESTAMP_TZ'),
+    ('REGION_NAME', 'STRING(16)'),
+    ('INSTANCE_STATE', 'STRING(16)'),
+    ('INSTANCE_NAME', 'STRING(256)')
+]
+
+# Define the columns for the Security Group Landing table
+AWS_SG_LANDING_TABLE_COLUMNS = [
+    ('RAW_DATA', 'VARIANT'),
```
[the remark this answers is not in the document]
cherrera2001: done

src/connectors/aws_config.py
```diff
@@ -0,0 +1,248 @@
...
+        'newStage': 'finalized',
+        'newMessage': (
+            f"Please add this SQS Queue ARN to the bucket event notification "
+            f"channel for all object create events: {sqs_arn}"
+        )
+    }
+
+
+def test(base_name):
+    yield db.fetch(f'ls @DATA.{base_name}_STAGE')
```
andrey-snowflake: rm or make it adhere to —

    type TestResult = {
        success: bool,
        description: string,
    }

please

src/connectors/aws_asset_ingest.py
```diff
@@ -0,0 +1,357 @@
...
+    return elbs
+
+
+def datetime_serializer(obj):
+    if isinstance(obj, datetime):
+        return obj.isoformat()
+    raise TypeError(f"Type {type(obj)} not serializable")
+
+
+def test(name):
```
andrey-snowflake: ditto -- remove or add actual test logic, please

src/connectors/aws_config.py
```diff
@@ -0,0 +1,248 @@
...
+        'prompt': 'folder in S3 bucket where Config puts logs',
+        'default': 'AWSLogs/',
+        'required': True,
+    },
+    {
+        'type': 'str',
+        'name': 'aws_role',
+        'title': 'AWS Role',
+        'prompt': "ARN of Role we'll grant access to bucket",
+        'placeholder': 'arn:aws:iam::012345678987:role/my-test-role',
```
andrey-snowflake: `my-config-reader-role` since it's probably not for testing

PR #201, opened 2019-06-24
Modified Login Portal for Internal Use

jiahaoyuan:
Hi, I have modified the Login Portal. The install interface looks exactly the same and the customer would not even notice it. But for internal use, they should input "snowhouse.preprod2.int.snowflakecomputing.com:8084" 
for the first question. 

The .envs file shall look like this now: 

SNOWFLAKE_HOST
SH_ACCOUNT
SH_USER
SH_ROLE
SH_DATABASE
SH_WAREHOUSE
SH_REGION
SH_PORT
SH_PROTOCOL
PRIVATE_KEY
PRIVATE_KEY_PASSWORD

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,36 @@
+FROM python:3.7-slim-stretch
+
+WORKDIR /var/task
+
+RUN pip install --upgrade pip virtualenv pyflakes
+
+RUN mkdir -p ./snowalert
+RUN virtualenv ./snowalert/venv
+ENV PATH="/var/task/snowalert/venv/bin:${PATH}"
+
+COPY ./src ./snowalert/src
+
+# backend
+RUN apt-get update \
+ && apt-get install -y gcc linux-libc-dev \
+ && rm -rf /var/lib/apt/lists/* \
+ && PYTHONPATH='' pip install ./snowalert/src/ ./snowalert/src/webui/backend/ \
+ && apt-get purge -y --auto-remove gcc linux-libc-dev
+
+# frontend
+RUN apt-get update \
+ && apt-get install -y curl gnupg2 apt-transport-https \
+ && curl -sL https://deb.nodesource.com/setup_11.x | bash - \
+ && apt-get install -y nodejs \
+ && curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
+ && echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
+ && apt-get update \
+ && apt-get install -y yarn \
+ && cd ./snowalert/src/webui/frontend && yarn install && yarn build \
+ && rm -fr node_modules \
+ && apt-get purge -y --auto-remove curl gnupg2 apt-transport-https nodejs yarn
+
+# link frontend build into backend venv
+RUN ln -s $PWD/snowalert/src/webui/frontend ./snowalert/venv/lib/python3.7/
+
+CMD python ./snowalert/src/webui/backend/webui/app.py
--- a/src/runners/helpers/dbconfig.py
+++ b/src/runners/helpers/dbconfig.py
@@ -15,18 +15,26 @@ if ENV == 'test':
 else:
     tail = ''
 
-# database & account properties
-REGION = environ.get('REGION', "us-west-2")
-REGION_SUBDOMAIN_POSTFIX = '' if REGION == 'us-west-2' else f'.{REGION}'
-ACCOUNT = environ.get('SNOWFLAKE_ACCOUNT', '') + REGION_SUBDOMAIN_POSTFIX
 
-USER = environ.get('SA_USER', "snowalert") + tail
+
+# database & account properties
+# For customer: HOST = '', PORT = '443', PROTOCOL = 'https'
+# For internal use: REGION = ''
+HOST = environ.get('SNOWFLAKE_HOST', '') # '' for customer,
+PORT = environ.get('SH_PORT')
+PROTOCOL = environ.get('SH_PROTOCOL')
+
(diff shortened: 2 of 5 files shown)


src/runners/helpers/dbconfig.py
```diff
@@ -15,18 +15,26 @@ def random_string(n=10, alphabet=ascii_uppercase):
 else:
     tail = ''
 
+
+
 # database & account properties
-REGION = environ.get('REGION', "us-west-2")
+# For customer: HOST = '', PORT = '443', PROTOCOL = 'https'
+# For internal use: REGION = ''
```
andrey-snowflake: don't mention internal use in public repo

src/runners/helpers/dbconfig.py
```diff
@@ -15,18 +15,26 @@ def random_string(n=10, alphabet=ascii_uppercase):
 else:
     tail = ''
 
+
+
```
andrey-snowflake: don't add extra newlines that break PEP-8 standards

src/runners/helpers/dbconfig.py
```diff
@@ -15,18 +15,26 @@ def random_string(n=10, alphabet=ascii_uppercase):
...
+# For customer: HOST = '', PORT = '443', PROTOCOL = 'https'
+# For internal use: REGION = ''
+HOST = environ.get('SNOWFLAKE_HOST', '') # '' for customer,
+PORT = environ.get('SH_PORT')
+PROTOCOL = environ.get('SH_PROTOCOL')
+
+REGION = environ.get('SH_REGION', "us-west-2")
 REGION_SUBDOMAIN_POSTFIX = '' if REGION == 'us-west-2' else f'.{REGION}'
-ACCOUNT = environ.get('SNOWFLAKE_ACCOUNT', '') + REGION_SUBDOMAIN_POSTFIX
+ACCOUNT = environ.get('SH_ACCOUNT', '') if HOST else environ.get('SH_ACCOUNT', '') + REGION_SUBDOMAIN_POSTFIX
```
andrey-snowflake: keep this env var name the same, as it's the Snowflake account used for SnowAlert

src/runners/helpers/dbconfig.py
```diff
@@ -15,18 +15,26 @@ def random_string(n=10, alphabet=ascii_uppercase):
...
 
-USER = environ.get('SA_USER', "snowalert") + tail
+USER = environ.get('SH_USER', "snowalert") + tail
 PRIVATE_KEY_PASSWORD = environ.get('PRIVATE_KEY_PASSWORD', '').encode('utf-8')
 PRIVATE_KEY = b64decode(environ['PRIVATE_KEY']) if environ.get('PRIVATE_KEY') else None
 
-ROLE = environ.get('SA_ROLE', "snowalert") + tail
-WAREHOUSE = environ.get('SA_WAREHOUSE', "snowalert") + tail
-DATABASE = environ.get('SA_DATABASE', "snowalert") + tail
+ROLE = environ.get('SH_ROLE', "snowalert") + tail
```
andrey-snowflake: `SH_` is non-public, so let's keep using `SA_` for SnowAlert

src/scripts/install.py
```diff
@@ -140,11 +144,20 @@ def parse_snowflake_url(url):
     if len(c) == 1:
         account = c[0]
     else:
+        # for customers
         if path.endswith("snowflakecomputing.com"):
             account = c[0]
             region = c[1] if len(c) == 4 else 'us-west-2'
+        else:
+        # for internal use
```
andrey-snowflake: as mentioned in our earlier meeting, please don't differentiate between internal and external use

src/scripts/install.py
```diff
@@ -140,11 +144,20 @@ def parse_snowflake_url(url):
...
+            searchObj = re.search('(.+snowflakecomputing.com):([0-9]+)', path)
```
andrey-snowflake: since the port is optional, best to make a regex that accepts strings with and without it. use a tool like https://regex101.com/ and let me know if you need help!

src/scripts/install.py
```diff
@@ -363,20 +382,26 @@ def setup_authentication(jira_password, region, pk_passphrase=None):
     return private_key, pk_passphrase, jira_password, rsa_public_key
 
 
-def gen_envs(jira_user, jira_project, jira_url, jira_password, account, region,
+def gen_envs(jira_user, jira_project, jira_url, jira_password, account, region, host, port, protocol,
              private_key, pk_passphrase, aws_key, aws_secret, **x):
+    # for the .envs file being written, we redesigned the format,
+    # so that currently, host become the TOP argument
```
andrey-snowflake: what are you trying to clarify with this comment? older versions of the format don't seem pertinent to someone reading this code

src/scripts/install.py
```diff
@@ -363,20 +382,26 @@ def setup_authentication(jira_password, region, pk_passphrase=None):
...
+        ('SH_WAREHOUSE', WAREHOUSE),
+        ('SH_REGION', region or 'us-west-2'),
+        ('SH_PORT', port),
+        ('SH_PROTOCOL', protocol),
 
         ('PRIVATE_KEY', b64encode(private_key).decode("utf-8")),
         ('PRIVATE_KEY_PASSWORD', pk_passphrase),
     ]
 
+
```
andrey-snowflake: please avoid adding extra newlines. also, please set up your text editor to warn you about violating PEP-8.

Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM python:3.7-slim-stretch
```
andrey-snowflake: not sure what this file is doing in this PR, but happy to explain how we use Docker if you'd like to build for the files which are committed.

PR #204, opened 2019-06-26
collects aws account information to snowflake

greg-snowflake:
This is not a perfect solution, but it ingests a collection of data (account id, account alias, number of instances, or an error) to snowflake (and since it's in a try/catch block, if the insert fails for any reason it doesn't kill the rest of the ingest process).

I'd do more refactoring of the script itself (e.g. to use the db functions instead of constructing raw sql) but this script will be obsoleted with a connector soon and it doesn't seem like a productive use of time to fix this up more than this.

### DIFF
--- a/src/ingestion/ec2_describe_instances.py
+++ b/src/ingestion/ec2_describe_instances.py
@@ -7,7 +7,7 @@ import os
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from multiprocessing import Pool
-from runners.utils import groups_of, json_dumps
+from runners.utils import groups_of
 from runners.helpers import db
 
 import snowflake.connector
--- a/src/ingestion/ec2_describe_instances.py
+++ b/src/ingestion/ec2_describe_instances.py
@@ -7,12 +7,14 @@ import os
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from multiprocessing import Pool
-from runners.utils import groups_of
+from runners.utils import groups_of, json_dumps
+from runners.helpers import db
 
 import snowflake.connector
 
 INSTANCES_TABLE = os.environ['EC2_INSTANCE_LIST_SNOWFLAKE_TABLE_IDENTIFIER']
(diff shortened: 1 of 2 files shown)


src/ingestion/ec2_describe_instances.py
```diff
@@ -114,12 +116,18 @@ def get_data_worker(account):
                 instances.extend(region)
             except Exception as e:
                 print(f"ec2_describe_instances: account: {account[1]} exception: {e}")
+                db.insert(AWS_ACCOUNTS_INFORMATION_TABLE, values=[(datetime.datetime.utcnow(), account[0], account[1], None, json_dumps(e))])
                 return None
         instance_list = [json.dumps({**instance,"AccountId":account[0]}, default=str) for instance in instances]
+        try:
+            db.insert(AWS_ACCOUNTS_INFORMATION_TABLE, values=[(datetime.datetime.utcnow(), account[0], account[1], len(instance_list), None)])
```
andrey-snowflake: add newlines before column 118 pls

src/ingestion/ec2_describe_instances.py
```diff
@@ -114,12 +116,18 @@ def get_data_worker(account):
...
+        except Exception:
+            print('Failed to insert into AWS_ACCOUNT_INFORMATION table.')
         print(f"ec2_describe_instances: account: {account[1]} instances: {len(instance_list)}")
         return instance_list
     except Exception as e:
         print(f"ec2_describe_instances: account: {account[1]} exception: {e}")
+        db.insert(AWS_ACCOUNTS_INFORMATION_TABLE, values=[(datetime.datetime.utcnow(), account[0], account[1], None, json_dumps(e))])
```
andrey-snowflake: formatting how something is inserted (i.e. `json_dumps`) should be something this code does not concern itself with -- can we put it somewhere into db?

src/ingestion/ec2_describe_instances.py
greg-snowflake: resolved

src/ingestion/ec2_describe_instances.py
greg-snowflake: Seems reasonable. I added Exception to the tuple of things which are jsony in `db.insert()`, so it'll json_dumps them the same way it does when you pass it a dict, list, or tuple.

PR #206, opened 2019-07-03
Aws account connector

greg-snowflake:
tested in oz03309


src/connectors/aws_account.py
```diff
@@ -0,0 +1,100 @@
...
+    }
+]
+
+LANDING_TABLE_COLUMNS = [
+    ('raw', 'VARIANT'),
+    ('event_time', 'TIMESTAMP_LTZ'),
+]
+
+
+def get_org_client(session_name, src_role_arn, dest_role_arn, dest_external_id):
```
andrey-snowflake: this seems like it's doing something else until the very end -- can we move out the `.client('organizations')` and give this helper method a better name? it seems like something that might be more general than this connector, also? what's a good place to put it?

src/connectors/aws_account.py
```diff
@@ -0,0 +1,100 @@
...
+from runners.helpers import db
+from runners.helpers.dbconfig import ROLE as SA_ROLE
+
+import boto3
+import datetime
+
+CONNECTION_OPTIONS = [
+    {
+        'name': 'source_role_arn',
+        'title': "source role arn",
```
andrey-snowflake: capitalize and make these something more reasonable

src/connectors/aws_account.py
```diff
@@ -0,0 +1,100 @@
...
+    {
+        'name': 'destination_role_arn',
+        'title': "destination role arn",
+        'prompt': "destination role arn",
+        'type': 'str',
+        'required': True
+    },
+    {
+        'name': 'destination_role_external_id',
+        'title': 'destination_role_external_id',
```
andrey-snowflake: these should have `"` since they'll be displayed to humans and let's make them human readable

src/connectors/aws_account.py
```diff
@@ -0,0 +1,100 @@
...
+        'name': 'source_role_arn',
+        'title': "source role arn",
+        'prompt': "assumer role arn",
+        'type': 'str',
+        'required': True
+    },
+    {
+        'name': 'destination_role_arn',
+        'title': "destination role arn",
+        'prompt': "destination role arn",
```
andrey-snowflake: title should be different from prompt

src/connectors/aws_account.py
```diff
@@ -0,0 +1,100 @@
+"""AWS Account List
+Lists your AWS Accounts
```
andrey-snowflake: We say "Collects" on every other one, so let's do that here, and clarify that it's "in your Organization"?

greg-snowflake: resolved

src/connectors/aws_account.py
greg-snowflake: resolved

src/connectors/aws_account.py
greg-snowflake: fixed

src/connectors/aws_account.py
greg-snowflake: resolved

src/connectors/aws_account.py
greg-snowflake: resolved

PR #217, opened 2019-07-18
Adds Azure Subscription Inventory DC

andrey-snowflake:
tested by creating locally in WebUI 

image (3)

image (2)

and then running

    (.venv) $ python run.py dc AZURE_SUBSCRIPTION_DEFAULT_CONNECTION
    [6989] -- START DC AZURE_SUBSCRIPTION_DEFAULT_CONNECTION --
    [6989] RUN metadata recorded.
    [6989] -- END DC --

and verifying that the table now contains the subscriptions


src/connectors/azure_subscription.py
```diff
@@ -0,0 +1,112 @@
...
+        'title': "Client Secret",
+        'prompt': "Secret access key authenticating the client",
+        'type': 'str',
+        'secret': 'true',
+        'required': True
+    },
+]
+
+LANDING_TABLE_COLUMNS = [
+    ('created_at', 'TIMESTAMP_LTZ'),
```
greg-snowflake: Shouldn't this be EVENT_TIME and not CREATED_AT?

PR #221, opened 2019-07-19
environment variable REGION is used for both KMS encryption and also …

nagalakshmisreeram:
…to build snowflake url. It was not working for our scenario where KMS was in us-east-1 but using snowflake accoun in us-west-2 deployment.


andrey-snowflake:
merged into v1.8.2

PR #225, opened 2019-07-19
Records timestamp of snapshot creation as event_time

greg-snowflake:
tested in dev by running the UI locally and creating the connector, hooking it up to an S3 bucket in a sandbox account, and manually uploading config snapshots to the s3 bucket

### DIFF
--- a/src/connectors/aws_config.py
+++ b/src/connectors/aws_config.py
@@ -55,7 +55,7 @@ LANDING_TABLE_COLUMNS = [
     ('aws_region', 'STRING'),
     ('arn', 'STRING'),
     ('availability_zone', 'STRING'),
-    ('resource_creation_time', 'TIMESTAMP_LTZ(9)'),
+    ('resource_modification_time', 'TIMESTAMP_LTZ(9)'),
     ('resource_name', 'STRING'),
     ('resource_Id', 'STRING'),
     ('resource_type', 'STRING'),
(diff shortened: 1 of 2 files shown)


andrey-snowflake:
So as with the Azure, `event_time` is probably best set via the file or path name. Insertion time is traditionally called `created_on`

greg-snowflake:
"created_on" sounds misleading because I would expect that to be the
creation time of the resource being described, as opposed to the creation
time of the row itself.

andrey-snowflake:
just Googled real quick and, `created_on` for dates / `created_at` for datetimes are standard in most SQL databases and also Mongo — I think it was RoR or Django that started it, but it's definitely something most folks will recognize as the time the row was created.

regardless, here, what we want is the time the event was created by the AWS Config snapshot, so either the time the file was written, or the day that the file corresponds to?

greg-snowflake:
This was tested in dev by running the UI locally and creating the connector, hooking it up to an S3 bucket in a sandbox account, and manually uploading config snapshots to the s3 bucket.

src/connectors/aws_config.py
```diff
@@ -175,12 +176,13 @@ def finalize(connection_name):
...
 INSERT INTO {landing_table} (
-  raw, hash_raw, event_time, account_id, aws_region, resource_type, arn, availability_zone,
-  resource_creation_time, resource_name, resource_Id, relationships, configuration, tags
+  raw, hash_raw, event_time, configuration_item_capture_time, account_id, aws_region, resource_type, arn,
+  availability_zone, resource_creation_time, resource_name, resource_Id, relationships, configuration, tags
 )
 SELECT value raw
     , HASH(value) hash_raw
-    , value:configurationItemCaptureTime::TIMESTAMP_LTZ(9) event_time
+    , REGEXP_REPLACE(filename, '.+(\\\\d{{4}})(\\\\d{{2}})(\\\\d{{2}})T(\\\\d{{2}})(\\\\d{{2}})(\\\\d{{2}})Z.*', '\\\\1-\\\\2-\\\\3T\\\\4:\\\\5:\\\\6Z')::timestamp_ltz event_time
```
andrey-snowflake: what happens if the regex doesn't match?

greg-snowflake: regexp_replace returns the original subject if there's no match, so the statement would fail when it gets a bad timestamp.

PR #233, opened 2019-07-26
Tenable connector

greg-snowflake:
Tested in the Development environment by running the UI locally and providing credentials. This connector is structured in such a way that going forward, we should be able to extend it to create other tables similar to the AWS Asset Inventory connector.


src/connectors/tenable.py
```diff
@@ -0,0 +1,118 @@
...
+                      user.get('type', None),
+                      user.get('permissions', None),
+                      user.get('last_login_attempt', None),
+                      user.get('login_fail_count', None),
+                      user.get('login_fail_total', None),
+                      user.get('enabled', None),
+                      user.get('two_factor', None),
+                      user.get('lastlogin', None),
+                      user.get('uuid_id', None)) for user in users],
+              select="""column1, column2, PARSE_JSON(column3), column4, column5, column6, column7, column8,
```
andrey-snowflake: nothing wrong here, but curious to think through how this could be less error-prone, e.g. whether we can construct this `SELECT` clause from the value types, maybe?

src/connectors/tenable.py
```diff
@@ -0,0 +1,118 @@
...
+                        column14, PARSE_JSON(column15), to_timestamp(column16, 3)::timestamp_ltz, column17""")
+
+
+def create_user_table(connection_name, options):
+    table_name = f'data.TENABLE_USER_{connection_name}_CONNECTION'
+    token = options['token']
+    secret = options['secret']
+    comment = f"""
+---
+module: tenable
```
andrey-snowflake: Omer pointed out that `okta` may be better named `okta_logs` to separate it from `okta_configuration`. Would it make sense to make this distinction here, as well, and call it `tenable_settings`? In the future, we might have other settings other than user settings as other connection types in this module.

src/connectors/tenable.py
```diff
@@ -0,0 +1,118 @@
+"""Tenable
+Collects a variety of Tenable logs into tables
+"""
+from tenable.io import TenableIO
+from datetime import datetime
+
+from runners.helpers import db
+from runners.helpers.dbconfig import ROLE as SA_ROLE
+
+CONNECTION_OPTIONS = [
```
andrey-snowflake: So right now this will only have one option (and it should be the default), but might it make sense to make this ahead-of-time extensible with a `connection_type` with value `user`?

src/connectors/tenable.py
```diff
@@ -0,0 +1,118 @@
+"""Tenable
```
andrey-snowflake: > Tenable Settings

and

> Collects Tenable Settings using a Service User’s API Key

src/connectors/tenable.py
```diff
@@ -0,0 +1,118 @@
...
+                      user.get('two_factor', None),
+                      user.get('lastlogin', None),
+                      user.get('uuid_id', None)) for user in users],
+              select="""column1, column2, PARSE_JSON(column3), column4, column5, column6, column7, column8,
+                        column9, column10, to_timestamp(column11, 3)::timestamp_ltz, column12, column13,
+                        column14, PARSE_JSON(column15), to_timestamp(column16, 3)::timestamp_ltz, column17""")
+
+
+def create_user_table(connection_name, options):
+    table_name = f'data.TENABLE_USER_{connection_name}_CONNECTION'
```
andrey-snowflake: once the module is renamed, to be consistent with others, the landing table here should be `tenable_settings_{connection_name}_user_connection`

src/connectors/tenable.py
greg-snowflake: This is definitely a weak point of our insert wrapper, yeah; I agree that at some point it would be good to sit down and figure out how we can make this better so that you don't have to line up columns and make sure they agree with each other.

src/connectors/tenable.py
```diff
@@ -0,0 +1,124 @@
...
+CONNECTION_OPTIONS = [
+    {
+        'type': 'select',
+        'options': [
+            {'value': 'user', 'label': "Tenable Users"},
+        ],
+        'name': 'connection_type',
+        'title': "Settings Type",
+        'prompt': "The type of Tenable Settings information you are ingesting to Snowflake.",
+        'required': True
```
andrey-snowflake: add a `'default': 'user',` should give people two less clicks to set this up, since it's the only one for now

src/connectors/tenable.py
```diff
@@ -0,0 +1,124 @@
...
+    for user in users:
+        user['role'] = {
+            16: 'Basic',
+            24: 'Scan Operator',
+            32: 'Standard',
+            40: 'Scan Manager',
+            64: 'Administrator',
+        }.get(user['permissions'], 'unknown permissions {permissions}')
+
+    db.insert(table=f'data.{table_name}',
```
andrey-snowflake: per Python culture, it needs to be laughably bad before there are too many newlines. something about newspaper columns being 80 char wide. also, it seems better to have each value have its own line, and end with a trailing comma, i.e.  —

    db.insert(
        table=f'data.{table_name}',
        values=[(
            user.get('username', None),
            user.get('role', None),
            user,
            timestamp,
            user.get('uuid', None),
            user.get('id', None),
            user.get('user_name', None),
            user.get('email', None),
            user.get('type', None),
            user.get('permissions', None),
            user.get('last_login_attempt', None),
            user.get('login_fail_count', None),
            user.get('login_fail_total', None),
            user.get('enabled', None),
            user.get('two_factor', None),
            user.get('lastlogin', None),
            user.get('uuid_id', None),
        ) for user in users],
        select="""column1, column2, PARSE_JSON(column3), column4, column5, column6, column7, column8,
                  column9, column10, to_timestamp(column11, 3)::timestamp_ltz, column12, column13,
                  column14, PARSE_JSON(column15), to_timestamp(column16, 3)::timestamp_ltz, column17"""
    )

PR #237, opened 2019-07-29
[DC] Adds Flow logs Connector w/ protocol mapping table

greg-snowflake:
Tested by running locally and verifying that collection works properly, using the flow logs set up to record in our development environment.


andrey-snowflake:
Does this work with VPC Flow logs or any kind of Flow log?

PR #239, opened 2019-07-30
[Dispatcher] Adds assignee and area params to Jira Handler

greg-snowflake:


src/runners/handlers/jira.py
```diff
@@ -144,7 +149,7 @@ def bail_out(alert_id):
         log.error(e, f"Failed to update alert {alert_id} with handler status")
 
 
-def handle(alert, correlation_id, project=PROJECT):
+def handle(alert, correlation_id, project=PROJECT, assignee=None, custom_field=None):
```
greg-snowflake: There's probably a nice way to use kwargs here?

andrey-snowflake: You could, but it might break things unexpectedly (e.g. if extra params at in the alert definition, it's better to ignore them rather than trying to pass them into `create_jira_ticket` and throwing an exception).

andrey-snowflake:
tested? the commit titles are a little confusing, but title the entire thing

    [Dispatcher] Adds assignee and area params to Jira Handler

greg-snowflake:
This was tested in the development environment, creating tickets in our Dev Jira project.

PR #243, opened 2019-07-31
[DC] Adds G Suite Logs Connector

andrey-snowflake:
tested by
1. running WebUI locally against oz
2. creating connector using WebUI
3. adding debug lines and repeatedly running connector with 

```$ python ./runners/run.py dc GSUITE_LOGS_DEFAULT_LOGIN_CONNECTION```

output showed the expected number of each event being added for each account


andrey-snowflake:
moving to PR on v1.8.4

PR #257, opened 2019-08-05
[DC] Adds G Suite Logs Connector

andrey-snowflake:
tested by running in dev


src/connectors/gsuite_logs.py
```diff
@@ -0,0 +1,139 @@
...
+            """{"type": "service_account", "project_id": "my-gcp-project", """
+            """"private_key_id": "34d2d9933e3f4a9ee55b968758e5ca3c2b348221", """
+            """"private_key": "-----BEGIN PRIVATE KEY-----\n...-----END PRIVATE KEY-----\n", """
+            """... }"""
+        ),
+        'required': True,
+        'secret': True,
+    },
+    {
+        'type': 'list',
```
greg-snowflake: should swap 'type' and 'name' here so that the ordering is consistent with the other two

PR #258, opened 2019-08-06
[DC] Adds Data Connector for VPC Flow Logs

andrey-snowflake:
tested by running in dev against oz, following instructions, and then running 

    andrey#SNOWALERT@DELETEME.DATA>SELECT COUNT(*)
                                   FROM data.aws_vpc_flow_log_zxcv_connection;

    +----------+
    | COUNT(*) |
    |----------|
    |      625 |
    +----------+
    1 Row(s) produced. Time Elapsed: 0.197s


src/connectors/aws_flow_log.py
```diff
@@ -0,0 +1,224 @@
...
+                f'FROM @data.{base_name}_stage/)'
+            ),
+            replace=True,
+            autoingest=True,
+        ),
+        n=10,
+        sleep_seconds_btw_retry=1
+    )
+
+    db.execute(f'ALTER PIPE {pipe} REFRESH')
```
andrey-snowflake: this seems like it will work well if they just turned on flow logs, but if they have had it running for awhile, this will block for 60 minutes, before swallowing the error?

src/runners/helpers/db.py
```diff
@@ -384,14 +399,37 @@ def get_pipes(schema):
     return fetch(f"SHOW PIPES IN {schema}")
 
 
-def create_stage(name, url, prefix, cloud, credentials, file_format, replace=False, comment=''):
+def create_table_from_csv(name, columns, file_path, file_format, ifnotexists=False):
```
andrey-snowflake: sweet :-) the name is wrong,  tho -- this is creating the table from the columns arg, not the CSV, error-ing if the table exists, and then loading the CSV into it. maybe `create_table_and_upload_csv` would be a better name?

seems cleaning up the stage after the copy statement would be helpful, as well?

this is the sort of thing that could turn into a command line tool pretty easily with pyfire and a couple of default, e.g. —

    db upload-csv --table=data.somedata --file-path=data.csv

src/connectors/aws_flow_log.py
greg-snowflake: Do we have a way to resolve this issue? Config faces the same dilemma.

src/runners/helpers/db.py
greg-snowflake: I agree with the rename, but I'm not clear what you mean when you say "this is creating the table from the column arg [and] erroring if the table exists"; it'll error if the table a) exists and b) doesn't have the columns that the CSV expects, but the error will be a result of trying to load data in, not table creation. I'm not sure what a better way to handle this would be; you could "CREATE OR REPLACE" the table, but then we lose whatever data was in there originally; I feel like it's better to create if not exists and then if you get an existing table, put the data in there and error if it's wrong.

greg-snowflake: Renamed table and turned on temporary stage for the function.

Not clear what you mean about "erroring if the table exists"; the function should only error if a) the table exists and b) doesn't have the correct columns, in which case it seems safe to assume that it's not actually the right table and dropping it would have been a bummer.

src/connectors/aws_flow_log.py
andrey-snowflake: Not sure what's available, but if there is a way to async enqueue this and continue down the function, that would probably be ideal. If dropped SQS is still a problem, we could probably solve both by having a stored procedure that maintains a set of TASKs that refresh a subset of the PIPE, maybe one per account as we had elsewhere?

The way that would work for sure is to make this an hourly task, but that seems like it may be worse than having anything at all, as it would leave our biggest partners in a place where they have a partial refresh running every hour.

src/runners/helpers/db.py
andrey-snowflake: Push the change? Got it -- so the second time someone tries to create the table it'll just ignore it.

src/connectors/aws_flow_log.py
andrey-snowflake: How about just skipping it and telling users to run it manually?

greg-snowflake: That works for me; the only reason you need to refresh the pipe is to pick up old files. If people want to backfill their data, they can refresh; connectors are meant to ingest data created after the connection is initialized.

andrey-snowflake:
Rebased on v1.8.4 one more time + re-tested

PR #262, opened 2019-08-09
[DC] AWS Accounts Connector

greg-snowflake:
tested by running the UI locally and creating the connection in the dev environment, then running it individually with `python run.py dc AWS_ACCOUNTS_DEFAULT_CONNECTION`. 

Produces a columnar table with one row per AWS Account in the organization.


greg-snowflake:
Tested the final changes, confirmed to work. Merging.

PR #268, opened 2019-08-13
External table uses substring instead of regexp

greg-snowflake:


greg-snowflake:
External Tables must use substr instead of regexp

PR #270, opened 2019-08-14
[DC] GitHub Webhook via S3

alldoami:
Tested by creating the connector and filling the tables with data in Snowflake using "ALTER PIPE ... REFRESH". We were able to successfully see the data in the tables from our AWS GitHub bucket.


src/connectors/github_organizations.py
```diff
@@ -0,0 +1,343 @@
+"""AWS Config
+Collect Config logs from S3 using a privileged Role
```
edulop91: looks like this comment got copy/pasted from config

src/connectors/github_organizations.py
```diff
@@ -0,0 +1,343 @@
...
+        'prefix': 's3://',
+        'placeholder': 'my-test-s3-bucket',
+        'required': True,
+    },
+    {
+        'type': 'str',
+        'name': 'filter',
+        'title': 'Prefix Filter',
+        'prompt': 'The folder in S3 bucket where Github Organizations puts logs',
+        'default': 'AWSLogs/',
```
edulop91: We might want some other default prefix here. maybe `/` for now

src/connectors/github_organizations.py
```diff
@@ -0,0 +1,343 @@
...
+    IGNORE_UTF8_ERRORS = FALSE,
+    SKIP_BYTE_ORDER_MARK = TRUE
+"""
+
+LANDING_TABLE_COLUMNS = [
+    ('insert_id', 'NUMBER IDENTITY START 1 INCREMENT 1'),
+    ('insert_time', 'TIMESTAMP_LTZ(9)'),
+    ('raw', 'VARIANT'),
+    ('hash_raw', 'NUMBER'),
+    ('_modified', 'TIMESTAMP_LTZ(9)'),
```
edulop91: this col might've carried over from fivetran schema

src/connectors/github_organizations.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+STEP 2: For Role "{role}", add the following inline policy:
+
+{role_policy}
+"""
+
+
+def connect(connection_name, options):
+    base_name = f'AWS_CONFIG_{connection_name}_EVENTS'.upper()
```
edulop91: these base names reference aws config, update to reference github

src/connectors/github_organizations.py
```diff
@@ -0,0 +1,341 @@
...
+    stage = f'data.{base_name}_STAGE'
+    staging_table = f'data.{base_name}_STAGING'
+    landing_table = f'data.{base_name}_CONNECTION'
+
+    bucket = options['bucket_name']
+    prefix = options.get('filter', 'AWSLogs/')
```
edulop91: this default prefix is defined in a couple of places, maybe might be worth making a constant

src/connectors/github_organizations.py
```diff
@@ -0,0 +1,341 @@
...
+    , value:branches_url:VARIANT branches
+    , value:created_at::TIMESTAMP_LTZ(9) created_at
+    , value:updated_at::TIMESTAMP_LTZ(9) updated_at
+    , value:assignees_url::VARIANT assignee
+    , value:releases_url::VARIANT release
+    , value:members_url::VARIANT membership
+    , value:FILL IN ALERT
+    , value:FILL IN SCOPE
+    , value:members:VARIANT member
+    , value:requested_reviewers::VARIANT requested_reviewer
```
edulop91: `requested_reviewers` vs `requested_reviewer` something might be off there

src/connectors/github_organizations.py
```diff
@@ -0,0 +1,341 @@
...
+        n=10,
+        sleep_seconds_btw_retry=1
+    )
+
+    db.create_task(name=f'data.{base_name}_TASK', schedule='1 minute',
+                   warehouse=WAREHOUSE, sql=config_ingest_task)
+
+    db.execute(f"ALTER PIPE {pipe} REFRESH")
+
+    pipe_description = list(db.fetch(f'DESC PIPE {pipe}'))
```
edulop91: seems like the prefered way is to do `next()` instead of casting this to a list

src/connectors/github_organizations.py
```diff
@@ -0,0 +1,341 @@
...
+            autoingest=True,
+        ),
+        n=10,
+        sleep_seconds_btw_retry=1
+    )
+
+    db.create_task(name=f'data.{base_name}_TASK', schedule='1 minute',
+                   warehouse=WAREHOUSE, sql=config_ingest_task)
+
+    db.execute(f"ALTER PIPE {pipe} REFRESH")
```
edulop91: refresh can take a long time, we should to like the config and cloudtrail connectors where the refresh command is surfaces to the user if they want to backfill

src/connectors/github_organizations.py
alldoami: make it a local variable?

src/connectors/github_organizations.py
alldoami: I checked the json file again and it says requested_reviewers

infra/github_connector/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## GitHub Webhooks to S3
+
+One way to aggregate POST requests in GitHub is using the below archiver that sends these requests to an AWS
+API Gateway which invokes a lambda. After the lambda processes the request, it is then sent to AWS Firehose to be
+formatted to be put into S3. You can find the Webhooks code here: https://github.com/chanzuckerberg/cztack
+
+```
+module "archiver" {
+  source = "cztack/github-webhooks-to-s3"
```
edulop91:
```suggestion
  source = "github.com/chanzuckerberg/cztack//github-webhooks-to-s3?ref=XXX"
```
TODO: fill out that ref

infra/github_connector/README.md
```diff
@@ -0,0 +1,46 @@
...
+that contains the role and the ExternalId that you should verify are the correct values.
+
+The finalize function has a parameter called `connection_name` which should be the same value as the `connection_name`
+parameter for the connect function. The result of calling this function is a message that suggests to run:
+```
+ALTER PIPE data.GITHUB_ORGANIZATIONS_DEFAULT_EVENTS_PIPE REFRESH;
+```
+in SnowFlake.
+
+To do this in cmd, run `python` and `import github_organizations`. Then to call these functions, just use:
```
edulop91: we were only doing this temporarily since we were having troubles with the webui/retrying connectors -- we might have to rm this

src/connectors/github_organizations.py
```diff
@@ -0,0 +1,342 @@
...
+        'prompt': 'The folder in S3 bucket where Github Organizations puts logs',
+        'default': S3_BUCKET_DEFAULT_PREFIX,
+        'required': True,
+    },
+    {
+        'type': 'str',
+        'name': 'aws_role',
+        'title': 'Github Organizations Bucket Reader Role',
+        'prompt': "Role to be assumed for access to Github Organizations files in S3",
+        'placeholder': 'arn:aws:iam::012345678987:role/my-config-reader-role',
```
edulop91:
```suggestion
        'placeholder': 'arn:aws:iam::012345678987:role/my-github-reader-role',
```

src/connectors/github_organizations.py
```diff
@@ -0,0 +1,342 @@
...
+        )
+    }
+
+
+def finalize(connection_name):
+    base_name = f'GITHUB_ORGANIZATIONS_{connection_name}_EVENTS'.upper()
+    pipe = f'data.{base_name}_PIPE'
+    landing_table = f'data.{base_name}_CONNECTION'
+
+    DATE_REGEXP = r'.+(\d{4})(\d{2})(\d{2})T(\d{2})(\d{2})(\d{2})Z.*'.replace("\\", "\\\\")
```
edulop91: unused

src/connectors/github_organizations.py
```diff
@@ -0,0 +1,342 @@
...
+        ),
+        n=10,
+        sleep_seconds_btw_retry=1
+    )
+
+    db.create_task(name=f'data.{base_name}_TASK', schedule='1 minute',
+                   warehouse=WAREHOUSE, sql=config_ingest_task)
+
+    pipe_description = next(db.fetch(f'DESC PIPE {pipe}'), None)
+    if len(pipe_description) is None:
```
edulop91:
```suggestion
    if pipe_description is None:
```

src/connectors/__init__.py
```diff
@@ -25,6 +27,7 @@
     'azure_log': azure_log,
     'azure_subscription': azure_subscription,
     'azure_vm': azure_vm,
+    'github_organizations': github_organization,
```
edulop91:
```suggestion
    'github_organization': github_organization,
```

src/connectors/github_organization.py
```diff
@@ -0,0 +1,341 @@
...
+from runners.helpers import db
+from runners.helpers.dbconfig import WAREHOUSE
+
+S3_BUCKET_DEFAULT_PREFIX = ""
+
+CONNECTION_OPTIONS = [
+    {
+        'type': 'str',
+        'name': 'bucket_name',
+        'title': 'Github Organizations Bucket',
```
edulop91:
```suggestion
        'title': 'Github Organization Bucket',
```

src/connectors/github_organization.py
```diff
@@ -0,0 +1,341 @@
...
+        'prompt': 'The S3 bucket Github Organizations puts your logs',
```
edulop91:
```suggestion
        'prompt': 'The S3 bucket Github Organization puts your logs',
```

src/connectors/github_organization.py
```diff
@@ -0,0 +1,341 @@
...
+        'prefix': 's3://',
+        'placeholder': 'my-test-s3-bucket',
+        'required': True,
+    },
+    {
+        'type': 'str',
+        'name': 'filter',
+        'title': 'Prefix Filter',
+        'prompt': 'The folder in S3 bucket where Github Organizations puts logs',
```
edulop91:
```suggestion
        'prompt': 'The folder in S3 bucket where Github Organization puts logs',
```

src/connectors/github_organization.py
```diff
@@ -0,0 +1,341 @@
...
+        'default': S3_BUCKET_DEFAULT_PREFIX,
+        'required': True,
+    },
+    {
+        'type': 'str',
+        'name': 'aws_role',
+        'title': 'Github Organizations Bucket Reader Role',
```
edulop91:
```suggestion
        'title': 'Github Organization Bucket Reader Role',
```

src/connectors/github_organization.py
```diff
@@ -0,0 +1,341 @@
...
+        'prompt': "Role to be assumed for access to Github Organizations files in S3",
```
edulop91:
```suggestion
        'prompt': "Role to be assumed for access to Github Organization files in S3",
```

src/connectors/github_organization.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+STEP 2: For Role "{role}", add the following inline policy:
+
+{role_policy}
+"""
+
+
+def connect(connection_name, options):
+    base_name = f'GITHUB_ORGANIZATIONS_{connection_name}_EVENTS'.upper()
```
edulop91:
```suggestion
    base_name = f'GITHUB_ORGANIZATION_{connection_name}_EVENTS'.upper()
```

src/connectors/github_organization.py
```diff
@@ -0,0 +1,341 @@
...
+                        }
+                    }
+                ]
+            }, indent=4),
+        )
+    }
+
+
+def finalize(connection_name):
+    base_name = f'GITHUB_ORGANIZATIONS_{connection_name}_EVENTS'.upper()
```
edulop91:
```suggestion
    base_name = f'GITHUB_ORGANIZATION_{connection_name}_EVENTS'.upper()
```

src/connectors/github_organization.py
```diff
@@ -0,0 +1,341 @@
...
+    pipe = f'data.{base_name}_PIPE'
+    landing_table = f'data.{base_name}_CONNECTION'
+
+    DATE_ISO8601_BACKREFERENCES = r'\1-\2-\3T\4:\5:\6Z'.replace("\\", "\\\\")
```
edulop91: unused

src/connectors/github_organization.py
```diff
@@ -0,0 +1,341 @@
...
+            ),
+            replace=True,
+            autoingest=True,
+        ),
+        n=10,
+        sleep_seconds_btw_retry=1
+    )
+
+    db.create_task(name=f'data.{base_name}_TASK', schedule='1 minute',
+                   warehouse=WAREHOUSE, sql=config_ingest_task)
```
edulop91:
```suggestion
                   warehouse=WAREHOUSE, sql=github_organization_ingest_task)
```

infra/github_connector/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## GitHub Webhooks to S3
+
+One way to aggregate POST requests in GitHub is using the below archiver that sends these requests to an AWS
+API Gateway which invokes a lambda. After the lambda processes the request, it is then sent to AWS Firehose to be
+formatted to be put into S3. You can find the Webhooks code here: https://github.com/chanzuckerberg/cztack
+
+```
+module "archiver" {
+  source = "github.com/chanzuckerberg/cztack/github-webhooks-to-s3"
```
edulop91: Just created the new release with this code: https://github.com/chanzuckerberg/cztack/releases/tag/v0.19.0

edulop91:
```suggestion
  source = "github.com/chanzuckerberg/cztack//aws-acm-certgithub-webhooks-to-s3?ref=v0.19.0"
```

src/connectors/github_organization.py
```diff
@@ -0,0 +1,339 @@
+from json import dumps
```
andrey-snowflake: At the top of each connector is a docstring that gives  the connector a human-readable name and a description that appear in the WebUI and succinctly explain what kind of data the user is to expect.

src/connectors/github_organization.py
```diff
@@ -0,0 +1,339 @@
...
+    COMPRESSION = "AUTO",
+    ENABLE_OCTAL = FALSE,
+    ALLOW_DUPLICATE = FALSE,
+    STRIP_OUTER_ARRAY = TRUE,
+    STRIP_NULL_VALUES = FALSE,
+    IGNORE_UTF8_ERRORS = FALSE,
+    SKIP_BYTE_ORDER_MARK = TRUE
+"""
+
+LANDING_TABLE_COLUMNS = [
```
andrey-snowflake: Am I understanding right that this is describing a commit?

src/connectors/github_organization.py
```diff
@@ -0,0 +1,339 @@
...
+    base_name = f'GITHUB_ORGANIZATION_{connection_name}_EVENTS'.upper()
+    stage = f'data.{base_name}_STAGE'
+    staging_table = f'data.{base_name}_STAGING'
+    landing_table = f'data.{base_name}_CONNECTION'
+
+    bucket = options['bucket_name']
+    prefix = options.get('filter', S3_BUCKET_DEFAULT_PREFIX)
+    role = options['aws_role']
+
+    comment = f"""
```
andrey-snowflake: I know some of our older code still has this, but please use `utils.yaml_dump(...)`

infra/github_connector/README.md
edulop91: https://www.terraform.io/docs/modules/sources.html#modules-in-package-sub-directories

edulop91: >When the source of a module is a version control repository or archive file (generically, a "package"), the module itself may be in a sub-directory relative to the root of the package.
A special double-slash syntax is interpreted by Terraform to indicate that the remaining path after that point is a sub-directory within the package. For example:

src/connectors/github_organization.py
```diff
@@ -0,0 +1,339 @@
...
+        file_format=FILE_FORMAT
+    )
+
+    db.create_table(
+        name=staging_table,
+        cols=[
+            ('value', 'VARIANT'),
+            ('filename', 'STRING(200)')
+        ]
+    )
```
greg-snowflake: I don't think you need the staging table/pipe/task model of connector here. That structure gets used for Cloudtrail because cloudtrail logs need to be flattened before they can be usefully queried. It doesn't look like the github logs need to be flattened (or at least your ingest task doesn't flatten anything when it pulls them from the stream). 

This means you can simplify the connector by modifying the pipe to copy files from S3 into the landing table and then you don't need the stream, task, and extra table.

src/connectors/github_organization.py
```diff
@@ -0,0 +1,339 @@
...
+        target=f'data.{base_name}_staging'
+    )
+
+    # IAM change takes 5-15 seconds to take effect
+    sleep(5)
+    db.retry(
+        lambda: db.create_pipe(
+            name=pipe,
+            sql=(
+                f"COPY INTO data.{base_name}_staging(value, filename) "
```
greg-snowflake: You want this pipe definition to be essentially what the ingest task is.

src/connectors/github_organization.py
```diff
@@ -0,0 +1,339 @@
...
+                f"FROM (SELECT $1, metadata$filename FROM @data.{base_name}_stage/)"
+            ),
+            replace=True,
+            autoingest=True,
+        ),
+        n=10,
+        sleep_seconds_btw_retry=1
+    )
+
+    db.create_task(name=f'data.{base_name}_TASK', schedule='1 minute',
```
greg-snowflake: This can be removed.

src/connectors/github_organization.py
```diff
@@ -0,0 +1,339 @@
...
+    , value:forkee::VARIANT forkee
+    , value:milestone::VARIANT milestone
+    , value:key::VARIANT key
+    , value:project_column::VARIANT project_column
+    , value:status::VARCHAR(256) status
+    , value:avatar_url::VARCHAR(256) avatar_url
+FROM data.{base_name}_stream
+'''
+
+    db.create_stream(
```
greg-snowflake: This can be removed

src/connectors/github_organization.py
```diff
@@ -1,35 +1,42 @@
+"""GitHub Webhooks to S3
+Collecting Webhooks from GitHub to S3 using an API Gateway, a Lambda, and Firehose.
```
greg-snowflake: This doesn't seem accurate; the connector brings logs from S3 to Snowflake. There's a separate piece, not in this file, which does the API Gateway/Lambda/Firehose piece to bring the logs into S3.

src/connectors/github_organization.py
alldoami: What is describing a commit? The skip_byte_order_mark?

src/connectors/github_organization.py
```diff
@@ -0,0 +1,251 @@
...
+                    , value:key::VARIANT key
+                    , value:project_column::VARIANT project_column
+                    , value:status::VARCHAR(256) status
+                    , value:avatar_url::VARCHAR(256) avatar_url
+                FROM data.{base_name}_stream
+                '''
+        )
+    )
+
+    db.insert(
```
greg-snowflake: This db.insert() call seems to be floating; it'll only get called once as part of the connection finalization process, which is likely not what we want to do.

src/connectors/github_organization.py
```diff
@@ -0,0 +1,251 @@
...
+                    , value:build::VARIANT build
+                    , value:deployment_status::VARIANT deployment_status
+                    , value:deployment::VARIANT deployment
+                    , value:forkee::VARIANT forkee
+                    , value:milestone::VARIANT milestone
+                    , value:key::VARIANT key
+                    , value:project_column::VARIANT project_column
+                    , value:status::VARCHAR(256) status
+                    , value:avatar_url::VARCHAR(256) avatar_url
+                FROM data.{base_name}_stream
```
greg-snowflake: Tasks select from streams (or from tables); a pipe should select from a stage and insert into a table.

alldoami: We don't have the stage anymore correct?

src/connectors/github_organization.py
alldoami: Ya I wasn't quite sure how to use the insert function. What would I put in there? Because I'm trying to go off of how the Okta connector inserts, but there are some values that aren't related to this connector.

src/connectors/github_organization.py
greg-snowflake: Ah, we definitely need a stage.

For an example of what the code might look like, check out the aws_vpc_flow_log here: https://github.com/snowflakedb/SnowAlert/pull/260/files#diff-e05b0e58621a6d15bb1884cba631fb54

The basic case of ingesting files from S3 to Snowflake requires a stage, a landing table, and a pipe. The stage is an interface between Snowflake and the external location, and gives Snowflake a file-level view. The pipe operates on files, parses them, and inserts (or copies) the results to the landing table.

The more complicated case is the one for Cloudtrail and Config, where you have the stage, landing table, and pipe, but also a staging table, stream, and ingest task. That's for situations where the file-level view isn't sufficient for ingestion (for example, if one row in the file contains multiple event rows of your logs, e.g. the Records key in a Cloudtrail file). You can't do a flatten inside of a pipe definition, so you have the pipe copy files to the stage, and then do the flattening inside of a task operating on a stream tied to the pipe, and the task copies from the staging table to the landing table.

Does that make sense? I'm happy to jump on a quick call with you to go over the code in person if that would be productive, or if the model I'm describing isn't clear.

src/connectors/github_organization.py
greg-snowflake: What other values are you trying to insert? The Okta connector uses db.insert because it scrapes data from an API, then processes that data itself and loads it into snowflake; data is never staged or stored anywhere. Here, we have data stored in S3, so we don't need to just raw insert it ourselves; the pipe definition will serve the function of parsing data, doing minor transformations, and loading it into the table.

src/connectors/github_organization.py
greg-snowflake: I think Andrey is referring to the landing table itself; is each row of the table (and by extension, each event in the logs) a commit in github?

edulop91: https://developer.github.com/v3/activity/events/types/ there are a bunch of webhook types other than commits

andrey-snowflake: Got it. The comment feature on GitHub shows 3 lines before the one I am commenting on, so I was referring to the `LANDING_TABLE_COLUMNS`.

Do I understand right that we're landing all of the webhooks into a table that has the union of all of the properties in each of them? That seems not ideal, but OK, as long as we make this clear to someone looking at the connector (if possible), or when reading the readme.

src/connectors/github_organization.py
```diff
@@ -0,0 +1,251 @@
+"""GitHub Webhooks Connector from S3
+Already collected Webhooks in S3 being placed in Snowflake
```
andrey-snowflake: Looking at the existing Connectors, it would be better to stay consistent with "Collect {what} from {where} using {credential}" —

image

So in this case, maybe something like —

image

?

alldoami: Ok! Sounds good thanks

andrey-snowflake:
lgtm, @greg-snowflake if you could take one more time you might spot something I missed, it might save us worry in the future :-)

@alldoami if you could double-check a run-through with the WebUI (Greg can help show you how to get it running if you guys haven't) and keep an eye out for any polish that might be helpful

@edulop91 could you check that `docs/connectors/github_webhooks_s3.md` makes sense at a high level -- I tried to change it to what I understand the infra here to be, so if you see anything off please don't hesitate to fix

docs/connectors/github_webhooks_s3.md
```diff
@@ -0,0 +1,32 @@
...
+- a branch is deleted,
+- a deploy key is added/removed from a repository, or
+- a new download is created in GitHub.
+
+SnowAlert does not yet natively support webhooks, or any other internet-facing infrastructure. Thus, a separate
+component is necessary to collect GitHub webhooks to S3 before using this connector to Snowflake. One way to
+accomplish this is to use the below archiver to send webhooks to an AWS API Gateway which invokes an AWS Lambda.
+After the Lambda processes the request, it is then sent to AWS Firehose to be formatted and put into S3.
+
+```
```
edulop91:
```suggestion
```hcl
```

docs/connectors/github_webhooks_s3.md
```diff
@@ -0,0 +1,32 @@
+## Getting GitHub Webhooks into S3
+
+GitHub allows Organization or Repository admins to set up webhooks for delivery to endpoints. Webhooks can be
```
edulop91:
```suggestion
GitHub allows Organization or Repository admins to set up webhooks for delivery to endpoints. [Webhooks](https://developer.github.com/v3/activity/events/types/) can be
```

docs/connectors/github_webhooks_s3.md
```diff
@@ -0,0 +1,32 @@
...
+triggered when
+
+- a new commit is pushed
+- a branch is deleted,
+- a deploy key is added/removed from a repository, or
+- a new download is created in GitHub.
```
edulop91:
```suggestion
- a new download is created in GitHub,
- GitHub package vulnerability alerts, or
- [many others](https://developer.github.com/v3/activity/events/types/)
```

docs/connectors/github_webhooks_s3.md
```diff
@@ -0,0 +1,32 @@
+## Getting GitHub Webhooks into S3
+
+GitHub allows Organization or Repository admins to set up webhooks for delivery to endpoints. Webhooks can be
+triggered when
+
+- a new commit is pushed
```
edulop91:
```suggestion
- a new commit is pushed,
```

docs/connectors/github_webhooks_s3.md
```diff
@@ -0,0 +1,32 @@
...
+- a branch is deleted,
+- a deploy key is added/removed from a repository, or
```
edulop91:
```suggestion
- a deploy key is added/removed from a repository,
```

docs/connectors/github_webhooks_s3.md
```diff
@@ -0,0 +1,32 @@
...
+- a new download is created in GitHub.
+
+SnowAlert does not yet natively support webhooks, or any other internet-facing infrastructure. Thus, a separate
+component is necessary to collect GitHub webhooks to S3 before using this connector to Snowflake. One way to
+accomplish this is to use the below archiver to send webhooks to an AWS API Gateway which invokes an AWS Lambda.
```
edulop91:
```suggestion
accomplish this is to use the [below archiver](https://github.com/chanzuckerberg/cztack/tree/master/github-webhooks-to-s3#github-webhooks-to-s3) to send webhooks to an AWS API Gateway which invokes an AWS Lambda.
```

docs/connectors/github_webhooks_s3.md
```diff
@@ -0,0 +1,32 @@
...
+  owner   = "${var.owner}"
+  service = "${var.component}"
+
+  fqdn            = "..."
+  certificate_arn = "..."
+  route53_zone_id = "..."
+}
+```
+
+The firehose.tf also uses cztack/aws-s3-private-bucket which can be found in the cztack repo.github_organizations.
```
edulop91:
```suggestion
```

docs/connectors/github_webhooks_s3.md
```diff
@@ -0,0 +1,32 @@
...
+By using this, we can then send data from S3 to Snowflake easily using the connector created.
```
edulop91:
```suggestion
```

PR #278, opened 2019-08-22
[DC] Adds osquery Data Connector

andrey-snowflake:
needs re-testing but other than that should be good


greg-snowflake:
This was tested by running the UI in a dev environment and creating a connection, verifying that the connection works, and then creating a second connection that uses the existing stage from the first connection.

Issue #255, opened 2019-08-02
Data Connector for CrowdStrike Falcon


edulop91:
Have some code for this we might be able to contribute back.

PR #279, opened 2019-08-22
[DC] Crowdstrike devices via API

edulop91:
Fixes https://github.com/snowflakedb/SnowAlert/issues/255


src/connectors/crowdstrike_devices.py
```diff
@@ -0,0 +1,279 @@
...
+
+def connect(connection_name, options):
+    table_name = f'crowdstrike_devices_{connection_name}_connection'
+    landing_table = f'data.{table_name}'
+
+    client_id = options['client_id']
+    client_secret = options['client_secret']
+
+    comment = yaml_dump(
+        module='crowdstrike_devices')
```
greg-snowflake: I think this needs to be something like 

```
comment = yaml_dump(module='crowdstrike_devices', **options)
```

Otherwise, you're not passing the client_id and client_secret back to the ingest method when you call it; the options are parsed from the comment when the connector calls ingest().

src/connectors/crowdstrike_devices.py
```diff
@@ -0,0 +1,279 @@
...
+    ('SERVICE_PACK_MAJOR', 'NUMBER'),
+    ('BUILD_NUMBER', 'NUMBER'),
+    ('POINTER_SIZE', 'NUMBER'),
+    ('SITE_NAME', 'STRING'),
+    ('MACHINE_DOMAIN', 'STRING'),
+    ('OU', 'VARIANT'),
+]
+
+
+def get_col_transform(idx: int) -> str:
```
greg-snowflake: This section is great and I'd love to see it integrated into db.py somehow instead of just being a part of the connector!

src/connectors/crowdstrike_devices.py
```diff
@@ -0,0 +1,279 @@
...
+    table_name = f'crowdstrike_devices_{connection_name}_connection'
+    landing_table = f'data.{table_name}'
+
+    client_id = options['client_id']
+    client_secret = options['client_secret']
+
+    comment = yaml_dump(
+        module='crowdstrike_devices')
+
+    db.create_table(name=landing_table,
```
greg-snowflake: You probably want to grant SELECT and INSERT on the landing_table to SA_ROLE; it's not always the case that the role being used to run the connector will be the role that snowalert uses to run alert queries or the connection runner.

PR #280, opened 2019-08-22
[DC] Cisco Umbrella Connector

alldoami:


src/connectors/cisco_umbrella.py
```diff
@@ -0,0 +1,227 @@
...
+    {
+        'name': 'api_secret',
+        'title': "Cisco Umbrella API Secret",
+        'prompt': "Your Cisco Umbrella API Secret",
+        'type': 'str',
+        'secret': True,
+        'required': True,
+    },
+    {
+        'name': 'organizations_id',
```
greg-snowflake: Is the organization id considered a secret?

src/connectors/cisco_umbrella.py
```diff
@@ -0,0 +1,227 @@
...
+from runners.helpers import db, log
+from runners.helpers.dbconfig import ROLE as SA_ROLE
+
+from datetime import datetime
+
+import requests
+from .utils import yaml_dump
+
+PAGE_SIZE = 500
+UMBRELLA_URL_DEVICE_BY_PAGE = "https://management.api.umbrella.com/v1/organizations/ORGANIZATION_ID/roamingcomputers"
```
greg-snowflake: Is the url really the same for every organization?

src/connectors/cisco_umbrella.py
```diff
@@ -0,0 +1,227 @@
...
+    ('OS_VERSION', 'VARCHAR(256)'),
+    ('NAME', 'VARCHAR(256)'),
+    ('STATUS', 'VARCHAR(256)'),
+    ('ORIGIN_ID', 'NUMBER(38,0)'),
+    ('APPLIED_BUNDLE', 'NUMBER(38,0)'),
+    ('HAS_IP_BLOCKING', 'NUMBER(38,0)'),
+]
+
+
+def get_col_transform(idx: int) -> str:
```
greg-snowflake: Ideally the get_col_transform and related code would be integrated into db.py and then connectors use it from there, so that we're not duplicating code across multiple connectors. I think Eduardo is working on that?

PR #281, opened 2019-08-23
[DC] Azure Subscription Connector supports Gov cloud

greg-snowflake:


greg-snowflake:
This was tested in dev for both gov and commercial cloud, and verified to populate the tables properly.

PR #290, opened 2019-08-28
[DC] Meraki Device Connector

edulop91:


greg-snowflake:
This branch removes the Cisco Umbrella icon; the commit which does that should be removed.

src/connectors/meraki.py
```diff
@@ -0,0 +1,181 @@
...
+                values=[(
+                    timestamp,
+                    client,
+                    client.get('id','None'),
+                    client.get('mac','None'),
+                    client.get('description','None'),
+                    client.get('mdnsName','None'),
+                    client.get('dhcpHostname','None'),
+                    client.get('ip','None'),
+                    parse_number(client.get('vlan','None')),
```
greg-snowflake: Can't this just be `client.get('vlan', None)` instead of wrapping it in parse_number?

src/connectors/meraki.py
```diff
@@ -0,0 +1,181 @@
...
+    db.create_table(name=landing_table_device,
+                    cols=LANDING_TABLE_COLUMNS_DEVICE, comment=comment)
+    db.execute(f'GRANT INSERT, SELECT ON {landing_table_device} TO ROLE {SA_ROLE}')
+    return {
+        'newStage': 'finalized',
+        'newMessage': "Meraki ingestion tables created!",
+    }
+
+
+def parse_number(value):
```
greg-snowflake: This function seems misnamed; I can pass any non-numerical value and get that value back. It seems like all this does is assert that a value is not None, but we have the value and we can check that ourselves anyway; we don't need to wrap that in a function. What problem are we trying to solve with this function, or what question are we trying to answer about the value we pass in?

src/connectors/meraki.py
```diff
@@ -0,0 +1,181 @@
...
+    if value:
+        return value
+    return None
+
+
+def ingest(table_name_client, landing_table_device, options):
```
greg-snowflake: The connection runner works on a per-table basis; that is, it gets a table, gets the options for the table, and passes both into the ingest method if one exists. It doesn't support passing multiple landing tables in for a single ingest call.

src/connectors/meraki.py
edulop91: The API is returning empty strings which then error out when trying to parse as a number

greg-snowflake: Hmm, okay. I'll think about this a bit; I suspect there's a better way to handle this.

src/connectors/meraki.py
```diff
@@ -0,0 +1,181 @@
...
+    except HTTPError as http_err:
+        log.error(f"Error GET: url={url}")
+        log.error(f"HTTP error occurred: {http_err}")
+        raise http_err
+    log.debug(req.status_code)
+    return req.json()
+
+
+def connect(connection_name, options):
+    landing_table_client = f'data.meraki_{connection_name}_connection_client'
```
greg-snowflake: The table names have to end in _CONNECTION or they're not considered a connection table (i.e. the connection runner won't pick them up and call the ingest method on the module for the table). So you'd want these to be `f'data.meraki_{connection_name}_client_connection'` or '_device_connection'.

src/connectors/meraki.py
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+
+def parse_number(value):
+    if value:
+        return value
+    return None
+
+
+def ingest(table_name, options):
```
greg-snowflake: So your connect method creates two tables, both of which I think you want to be connection tables. But the ingest method populates both of them every time it gets called, which means that you're going to be duplicating data in your tables; you'll populate both tables when the connection runner finds the client_connection table, and then you'll populate both tables again a second later when the connection runner finds the device_connection table. 

For each call to the ingest method, it should populate one table: the table which ingest is called for.

src/connectors/meraki.py
```diff
@@ -0,0 +1,181 @@
...
+        "X-Cisco-Meraki-API-Key": f"{token}",
+    }
+    try:
+        log.debug(f"Preparing GET: url={url} with params={params}")
+        req = requests.get(url, params=params, headers=headers)
+        req.raise_for_status()
+    except HTTPError as http_err:
+        log.error(f"Error GET: url={url}")
+        log.error(f"HTTP error occurred: {http_err}")
+        raise http_err
```
andrey-snowflake: re-raise with just `raise`, that way the stack trace is maintained rather than being reset to this point

src/connectors/meraki.py
andrey-snowflake: not sure what the client value is, but agreed this looks buggy. right now it does:

    client.get('vlan', 'None') or None

are you sure this is what you want?

src/connectors/meraki.py
```diff
@@ -0,0 +1,181 @@
...
+                    client,
+                    client.get('id','None'),
+                    client.get('mac','None'),
+                    client.get('description','None'),
+                    client.get('mdnsName','None'),
+                    client.get('dhcpHostname','None'),
+                    client.get('ip','None'),
+                    parse_number(client.get('vlan','None')),
+                    client.get('switchport','None'),
+                    parse_number(client.get('usage','None').get('sent','None')),
```
andrey-snowflake: You can't `.get` on a string (i.e. `'None'`), so this will raise an exception if `'usage' not in client`

src/connectors/meraki.py
```diff
@@ -0,0 +1,181 @@
+"""Meraki
```
andrey-snowflake: Let's make this name a little more descriptive -- there are a lot of things we could potentially collect from Meraki, and the titles should  be distinct.

src/connectors/meraki_devices.py
```diff
@@ -0,0 +1,194 @@
...
+                serial_number = device['serial']
+                
+                try:
+                    clients = get_data(f"https://api.meraki.com/api/v0/devices/{serial_number}/clients", api_key)
+                except requests.exceptions.HTTPError as e:
+                    log.error(f"{network} not accessible, ")
+                    log.error(e)
+                    continue
+                
+                db.insert(
```
andrey-snowflake: sorry I wasn't more clear on this. could you please replace this db.insert expression with --


                vlan = client.get('vlan') or None
                usage_sent = client.get('usage', {}).get('sent') or None
                usage_recv = client.get('usage', {}).get('recv') or None
                db.insert(
                    landing_table,
                    values=[(
                        timestamp,
                        client,
                        client.get('id', None),
                        client.get('mac', None),
                        client.get('description', None),
                        client.get('mdnsName', None),
                        client.get('dhcpHostname', None),
                        client.get('ip', None),
                        client.get('switchport', None),
                        vlan,
                        usage_sent,
                        usage_recv,
                        serial_number,
                    ) for client in clients],
                    select=db.derive_insert_select(LANDING_TABLE_COLUMNS_CLIENT),
                    columns=db.derive_insert_columns(LANDING_TABLE_COLUMNS_CLIENT)
                )

I think it will do the same thing but more succinctly.

to clarify `x or y` will return `y` if `x`  if "falsy", i.e. if `bool(x) is False`, and `dict`'s `.get` returns None as the default value if the key is not present

andrey-snowflake: also, are you sure `vlan` needs to be treated differently than `ip` or `switchport`? this seems off.

src/connectors/meraki_devices.py
```diff
@@ -0,0 +1,194 @@
...
+            log.error(e)
+            continue
+
+        if (ingest_type == 'device'):
+            db.insert(
+                landing_table,
+                values=[(
+                    timestamp,
+                    device,
+                    device.get('serial',None),
```
andrey-snowflake: please remove all of these `,None`, as they're vestigial (`None` is the default value when key is not present), and per pep8, the comma in param lists needs to be followed by a space or newline.

src/connectors/meraki_devices.py
```diff
@@ -0,0 +1,194 @@
...
+
+def connect(connection_name, options):
+    landing_table_client = f'data.meraki_devices_{connection_name}_connection_client'
+    landing_table_device = f'data.meraki_devices_{connection_name}_connection_device'
+    options['network_id_whitelist'] = options.get('network_id_whitelist', '').split(',')
+
+    comment = yaml_dump(module='meraki_devices', **options)
+
+    db.create_table(name=landing_table_client,
+                    cols=LANDING_TABLE_COLUMNS_CLIENT, comment=comment)
```
andrey-snowflake: add a newline before "comment" please to adhere to PEP8

andrey-snowflake:
@edulop91 could you give me push access to this repo so I can help push minor cleanup a little faster than explaining it in comments? I'll still comment on major things that have value to see but whitespace stuff seems it might become tedious.

src/connectors/meraki_devices.py
andrey-snowflake: Got it, interesting. If you're sure that the others never return `''` then lgtm. Adding a comment like --

                    client.get('vlan') or None,  # vlan sometimes set to ''

might help to future folks, as well

src/connectors/meraki_devices.py
```diff
@@ -0,0 +1,194 @@
...
+Collect Meraki Device information using an API Token
+"""
+
+from runners.helpers import log
+from runners.helpers import db
+from runners.helpers.dbconfig import ROLE as SA_ROLE
+
+from datetime import datetime
+
+import snowflake
```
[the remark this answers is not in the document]
kuannie1: Do you know if flake8 works for this purpose?

src/connectors/meraki_devices.py
```diff
@@ -0,0 +1,194 @@
...
+    db.execute(f'GRANT INSERT, SELECT ON {landing_table_device} TO ROLE {SA_ROLE}')
+    return {
+        'newStage': 'finalized',
+        'newMessage': "Meraki ingestion tables created!",
+    }
+
+
+def ingest(table_name, options):
+    ingest_type = ''
+    if (table_name.endswith('CONNECTION_CLIENT')) or (table_name.endswith('connection_client')):
```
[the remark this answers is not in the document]
kuannie1: I think I know the answer already, but does the table name have to be all-caps? (I'm guessing yes)

src/connectors/meraki_devices.py
```diff
@@ -0,0 +1,194 @@
...
+        log.error(f"HTTP error occurred: {http_err}")
+        raise
+    log.debug(req.status_code)
+    return req.json()
+
+
+def connect(connection_name, options):
+    landing_table_client = f'data.meraki_devices_{connection_name}_connection_client'
+    landing_table_device = f'data.meraki_devices_{connection_name}_connection_device'
+    options['network_id_whitelist'] = options.get('network_id_whitelist', '').split(',')
```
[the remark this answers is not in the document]
kuannie1: Would the special UI element parse the user input as a list? When we tested this particular input, we got it as a string, like `'id1, id2, id3'`.

src/connectors/meraki_devices.py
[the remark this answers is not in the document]
greg-snowflake: It depends. If you're working with the string in python, then yes; python is case-sensitive and so the strings '_connection' and '_CONNECTION' are not identical.

If you're working with the table in sql, then no; sql is generally case insensitive (more precisely, it will uppercase everything you say unless you explicitly tell it not to) and it will view `select * from data.meraki_connection` the same as `SELECT * FROM DATA.MERAKI_CONNECTION`. 

Note that you can get around this by putting something in double quotes inside sql; that is, 

`select * from "DATA"."MERAKI_CONNECTION"` is not the same as `select * from "data"."meraki_connection"`; one of those will look for a table with an uppercase name and one will look for a table with a lowercase name.

Does that answer your question? Happy to clarify if the above is confusing!

src/connectors/meraki_devices.py
kuannie1: I've never seen that occur with usage:recv or usage:sent, but I think it'll be best if we used the same approach for those instances too. Thank you for this feedback!

src/connectors/meraki_devices.py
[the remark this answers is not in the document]
kuannie1: would it be annoying if I typed:
`ingest_type = 'client' if (table_name.endswith('_CLIENT_CONNECTION')or table_name.endswith('_client_connection')) else 'device'` ?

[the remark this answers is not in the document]
greg-snowflake: What situation are you trying to protect against? Snowflake defaults to uppercase names, so you have to explicitly specify that you want an object to have a lowercase name by putting it in double-quotes. We don't do that in the connectors - do you have reason to think that somebody might create a connection table manually and explicitly name it "meraki_client_connection", and still want the connector to work with it?

src/connectors/meraki_devices.py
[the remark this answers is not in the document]
kuannie1: https://github.com/snowflakedb/SnowAlert/pull/306

kuannie1:
@greg-snowflake @andrey-snowflake: I think I'm done making changes. How does this look now? Given my pull request https://github.com/snowflakedb/SnowAlert/pull/306

src/connectors/meraki_devices.py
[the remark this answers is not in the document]
kuannie1: Ohh, I think I understand. Thank you for clarifying! I'm not worried about lower-case confusion anymore.

src/connectors/meraki_devices.py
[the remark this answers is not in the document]
andrey-snowflake: Perfect, thanks.

Err, had a bit of a think-o. The "return" type I was thinking of was of the expression

    options.get('network_id_whitelist', '').split(',')

but since it splits the empty string, it's always a list.

src/connectors/meraki_devices.py
andrey-snowflake: sgtm

PR #294, opened 2019-09-04
[DC] Assetpanda Connector

kuannie1:


src/connectors/assetpanda.py
```diff
@@ -0,0 +1,173 @@
...
+    ('INSERT_ID', 'NUMBER IDENTITY START 1 INCREMENT 1'),
+    ('RAW', 'VARIANT'),
+    ('ID', 'VARCHAR(256)'),
+    ('INSERT_AT', 'TIMESTAMP_LTZ(9)')
+]
+
+### Helper Functions ###
+
+# Retrieve the values needed from the results objects
+def get_list_objects_and_total_from_get_object(result: dict) -> (list, int):
```
greg-snowflake: This looks like it returns a tuple, not a dict.

src/connectors/assetpanda.py
```diff
@@ -0,0 +1,173 @@
...
+### Main Functions ###
+
+def get_data(token: str, url: str, params: dict = {}) -> dict:
+    headers: dict = {
+        "Authorization": f"Bearer {token}"
+    }
+    try:
+        log.debug(f"Preparing GET: url={url} with params={params}")
+        req = requests.get(url, params=params, headers=headers)
+        print("req: ", req)
```
greg-snowflake: should be log.debug or removed, not print

src/connectors/assetpanda.py
```diff
@@ -0,0 +1,173 @@
...
+        # replace every key "field_NO" by the value of the clear_field["field_NO"]
+        list_object_without_field_id = replace_device_key(list_object, clear_fields)
+
+
+        db.insert(
+            landing_table,
+            values=[(
+                entry,
+                entry.get('id', None),
+                datetime.utcnow()
```
greg-snowflake: This is going to give every page you insert a slightly different insert time. This will make it a bit challenging to get all of the latest results; you have to select where the insert_time is within a certain range and make sure your range doesn't include any results from the previous scan. If you mark an insert_time before you get into the `while` loop, you can use that insert_time for each page and then select from the table where insert_time = max(insert_time).

src/connectors/assetpanda.py
```diff
@@ -0,0 +1,174 @@
...
+        "limit": PAGE_SIZE,
+    }
+
+    total_object_count = 0 
+
+    insert_time = datetime.utcnow()
+
+    while params['offset'] <= total_object_count:
+
+        log.debug("total_object_count: ", total_object_count)
```
greg-snowflake: Do we need to log this information every time we go through the loop? It's not changing, so this is going to produce a log line for every loop with the same information; this seems less than ideal.

greg-snowflake:
Just two things left: first, can you add an assetpanda icon to src/webui/frontend/public/icons/connectors?

And secondly, can you comment on the PR describing the testing that you did to verify that it works? This can be something like "Ran the webui locally, created the connection, ran the ingest function, verified that data was loaded correctly"

src/connectors/assetpanda.py
```diff
@@ -0,0 +1,185 @@
...
+    """Retrieve the values needed from the results objects"""
+    try:
+        list_object: list = result["objects"]
+        total_object_count: int = result["totals"]["objects"]
+    except BaseException:
+        raise AttributeError("error format result")
+    return (list_object, total_object_count)
+
+
+def reduce_fields(accumulators: dict, field: dict) -> str:
```
andrey-snowflake: might this be better named "accumulated_value" or "result"? it's a single value that's passed through repeated calls of the reducer.

greg-snowflake:
It looks like you've somehow picked up a file and added it to the branch - src/baseline_modules/temporal_by_day/temporal_by_day.R.

This might just be the result of changes to the base branch; can you try rebasing this branch onto 1.8.6 for me to see if that resolves the issue?

src/baseline_modules/temporal_by_day/temporal_by_day.R
```diff
@@ -0,0 +1,101 @@
+require('dplyr')
```
greg-snowflake: This file shouldn't be a part of this PR!

PR #295, opened 2019-09-05
[DC] Nginx Connector

greg-snowflake:


greg-snowflake:
This was tested by running the webui locally and creating the connector. Note that it requires a specific folder layout for the nginx access and error logs (i.e. you specify a folder which contains two folders, one called 'access' and one called 'error', each of which contain their respective logs). This layout needs to be communicated to the user and configured by them for this connector to work.

src/connectors/nginx.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def connect(connection_name, options):
+    table_name = f'nginx_{connection_name}_connection'
+    log_landing_table = f'data.{table_name}'
+    error_landing_table = f'data.nginx_{connection_name}_error_connection'
+    prefix = ''
+    bucket_name = ''
+
+    for i in options:
```
andrey-snowflake: I think options is always a dict so this is more clear as

    if '' in options:
        del options['']

src/connectors/nginx.py
```diff
@@ -0,0 +1,290 @@
...
+    db.create_table(
+        name=error_landing_table,
+        cols=ERROR_LANDING_TABLE_COLUMNS,
+        comment=yaml_dump(module='nginx', **options)
+    )
+    db.execute(f'GRANT INSERT, SELECT ON {error_landing_table} TO ROLE {SA_ROLE}')
+
+    stage = options.get('existing_stage')
+    if stage:
+        stage_name = stage
```
andrey-snowflake: should this also clear out the other options as it does in the ldap connector?

src/connectors/nginx.py
```diff
@@ -0,0 +1,290 @@
...
+                    }
+                ]
+            }, indent=4),
+        )
+    }
+
+
+def finalize(connection_name):
+    base_name = f'nginx_{connection_name}'
+    options = yaml.load(next(db.fetch(f"SHOW TABLES LIKE '{base_name}_connection' IN data"))['comment'])
```
andrey-snowflake: to make this more clear split out

    table = next(db.fetch(f"SHOW TABLES LIKE '{base_name}_connection' IN data"))

src/connectors/nginx.py
```diff
@@ -0,0 +1,290 @@
...
+    error_pipe = f'data.{base_name}_error_pipe'
+
+    # IAM change takes 5-15 seconds to take effect
+    sleep(5)
+    db.retry(
+        lambda: db.create_pipe(
+            name=pipe,
+            sql=(
+                f'COPY INTO data.{base_name}_connection '
+                f"FROM (SELECT PARSE_JSON($1), HASH($1), regexp_substr($1:request::string, '.*request_?[iI]+d=([^\\\\&\\\\s]+)', 1, 1, 'e')"
```
andrey-snowflake: where is this from? how sure can we be it all works?

should I manually review this?

src/connectors/nginx.py
```diff
@@ -0,0 +1,290 @@
...
+            file_format=db.TypeOptions(type='JSON')
+        )
+
+    stage_props = db.fetch_props(
+        f'DESC STAGE {stage_name}',
+        filter=('AWS_EXTERNAL_ID', 'SNOWFLAKE_IAM_USER', 'AWS_ROLE', 'URL')
+    )
+
+    if prefix == '':
+        prefix = '/'.join(stage_props['URL'].split('/')[3:-1])
```
andrey-snowflake: cleanest would be regex, but cleaner to read if you split out

    url_parts = stage_props['URL'].split('/')

src/connectors/nginx.py
greg-snowflake: This is from the pipe definition that we're currently using to ingest nginx logs from s3 into snowhouse.

src/connectors/nginx.py
greg-snowflake: That doesn't have the same effect.

`>>> c = {'foo': '', 'bar': 'baz'}
>>> if '' in c:
... 	del c['']
...
>>> c
{'foo': '', 'bar': 'baz'}`

src/connectors/nginx.py
greg-snowflake: Yes. I wrote this one before I wrote that one, but I should bring that functionality over here as well. It means that if you provide an existing stage, you don't have to clear out the defaults from fields which provide one; it'll get automatically discarded (because we can derive everything we need from the existing stage and it's better to do that than ask users to remember what the stage says and risk them giving us the wrong information).

greg-snowflake: Looking at the code, we actually already do this; those values are initialized to empty strings at lines 108 and 109, and only populated with the options if we don't get an existing stage. And then the only time we use the AWS_ROLE option is when we're creating a new stage; the rest of the time, we get the role from the stage itself. So I think this one is actually fine as-is.

src/connectors/nginx.py
greg-snowflake: done

src/connectors/nginx.py
greg-snowflake: Turned this into a regex! you're correct, it's cleaner.

src/connectors/nginx.py
andrey-snowflake: `for i in option` iterates over the keys, not the values. if you're looking to remove empty values, it's probably clearest to re-construct the dict instead of mutating it, i.e.

    d = {k: v for k, v in d.items() if v != ''}

src/connectors/nginx.py
andrey-snowflake: sounds good. to make it a little more standard / clear, could we use the idiom I just changed `osquery_logs` to, i.e.

    stage_name = options.get('existing_stage')
    if not stage_name:
        ...

PR #297, opened 2019-09-05
[DC] Changed Cisco Umbrella Table Name

alldoami:
To indicate roaming clients


greg-snowflake:
Note that changes to existing connectors need to have migration steps included in a migration file in /src/migrations

PR #299, opened 2019-09-06
[DC] Fixed Cisco Umbrella Connector organization_id parsing

alldoami:


src/connectors/cisco_umbrella.py
```diff
@@ -104,7 +105,6 @@ def ingest(table_name, options):
     landing_table = f'data.{table_name}'
     timestamp = datetime.utcnow()
 
-    organizations_id = options['organization_id']
```
edulop91: @alldoami  we might still need this

src/connectors/cisco_umbrella.py
```diff
@@ -85,6 +85,7 @@ def get_data(organization_id: str, key: str, secret: str, params: dict = {}) ->
 def connect(connection_name, options):
     table_name = f'cisco_umbrella_{connection_name}_connection'
     landing_table = f'data.{table_name}'
+    options['organization_id'] = int(options['organization_id'])
```
edulop91: :+1:

greg-snowflake: Does this break backwards compatibility with a connector set up before this patch? Or do the options get treated the same once they get yaml-ized and stored in the table comment?

Just double checking.

andrey-snowflake: yaml treats strings and int differently, just like JSON does, so the comments in currently deployed connectors should be updated. don't think anyone else is using this one just yet, but in future we should include this in migration notes.

src/connectors/cisco_umbrella.py
```diff
@@ -33,7 +33,7 @@
         'name': 'organization_id',
         'title': "Cisco Umbrella Organization Id",
         'prompt': "Your Cisco Umbrella Organization Id",
-        'type': 'str',
+        'type': 'int',
```
andrey-snowflake: going to add this to the general runner here as we do with JSON: https://github.com/chanzuckerberg/SnowAlert/blob/v1.8.6/src/runners/connectors_runner.py#L43-L44

andrey-snowflake:
re-created under #301

PR #300, opened 2019-09-06
[DC] LDAP connector

greg-snowflake:
Tested by running the UI locally and creating the connection, verifying that the files are ingested properly, and then dropping everything but the stage and creating the connection again using the existing stage flow.


src/connectors/ldap.py
```diff
@@ -0,0 +1,219 @@
...
+        'prompt': "The S3 bucket where the LDAP Logs are collected",
+        'prefix': "s3://",
+        'placeholder': "my-ldap-log-s3-bucket",
+    },
+    {
+        'type': 'str',
+        'name': 'prefix',
+        'title': "Prefix Filter",
+        'prompt': "The folder in LDAP Log Bucket where logs are collected",
+        'default': "ldap/",
```
andrey-snowflake: how common is default? is it something only we have?

src/connectors/ldap.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+    }
+
+
+def finalize(connection_name):
+    base_name = f'ldap_{connection_name}'
+    pipe = f'data.{base_name}_pipe'
+    table = next(db.fetch(f"SHOW TABLES LIKE '{base_name}_connection' IN data"))
+    options = yaml.load(table['comment'])
+    stage = options.get('existing_stage', f'data.{base_name}_stage')
```
andrey-snowflake: I think in order for this to work, you have to add the  `existing_stage` to the `yaml_dump` call above?

src/connectors/__init__.py
```diff
@@ -12,6 +12,7 @@
 from . import tenable_settings
 from . import crowdstrike_devices
 from . import cisco_umbrella
+from . import ldap
```
andrey-snowflake: I remember regretting the `okta` without `_log` and we have it elsewhere, so might it make more sense to call this `ldap_logs`? no super strong intuition here, but t might be separate from `ldap_settings` or `ldap_accounts`?

src/connectors/ldap.py
greg-snowflake: IIRC ldap logs are ingested to S3 via FluentD, so there's not really a meaningful default to put here; it's what we set up when we configured fluentd.

src/connectors/ldap.py
greg-snowflake: Ah, good catch; the stage I was using adhered to the 'data.{base_name}_stage' format, which is why this made it through testing.

src/connectors/__init__.py
greg-snowflake: It was azure.py that we regretted (okta.py still exists), but your point is well taken.

PR #304, opened 2019-09-09
Add Rule: AWS S3 Bucket Open ACL

omer-snowflake:
edulop91 please test this against your data. I'll add you as a project contributor so that we can @ mention each other for rule creation, please let me know who else to add.

### DIFF
--- a/rules/AWS_S3_Bucket_Open_ACL.sql
+++ b/rules/AWS_S3_Bucket_Open_ACL.sql
@@ -1,0 +1,57 @@
+CREATE OR REPLACE VIEW rules.aws_s3_bucket_open_acl_violation_query COPY GRANTS
+  COMMENT='S3 Bucket ACLs can grant or deny access to specific entities. This rule checks for public bucket access.
+  @id 6rn7ad2y7tv
+  @tags aws, s3, scout2'
+AS
+SELECT OBJECT_CONSTRUCT(
+         'account_id', raw:"awsAccountId",
+         'resource_type', raw:"resourceType",
+         'name', raw:"resourceName",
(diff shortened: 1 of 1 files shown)


andrey-snowflake:
@edulop91 looks like we accidentally changed up v1.8.6 under this PR and made it look like something else -- could you take another look when you get the chance and get back to us with thoguhts and/or questions?

rules/AWS_S3_Bucket_Open_ACL.sql
```diff
@@ -0,0 +1,57 @@
+CREATE OR REPLACE VIEW rules.aws_s3_bucket_open_acl_violation_query COPY GRANTS
+  COMMENT='S3 Bucket ACLs can grant or deny access to specific entities. This rule checks for public bucket access.
+  @id 6rn7ad2y7tv
```
alldoami: What is this id syntax? What is it for?

omer-snowflake: The id in the comment is for the UI to use. Our BI also references it. Having the identifier in their let's us avoid pulling from the database for that info. Value should be the same as the query_id.

edulop91: is this just a random value?

omer-snowflake: Yup, exactly. Clicking on the +Query button in the Alerts or Violations section will generate boilerplate rule with a unique value in there.

andrey-snowflake: It's required for BI because of a permissions limitation on our end -- view-only roles can read the comment of a view but not the SQL.

PR #316, opened 2019-09-12
Change AWS Inventory Connection Identifier Name

alldoami:
Was confusing before what this entailed and had to look at the source code to see exactly what it was used for. I think it would be more clear if you just asked for the whole table name because then the user would then know that table must exist in order for the inventory table to be created.

### DIFF
--- a/src/connectors/aws_inventory.py
+++ b/src/connectors/aws_inventory.py
@@ -98,9 +98,9 @@ CONNECTION_OPTIONS = [
         'name': 'accounts_connection_name',
         'title': "AWS Accounts Table Name",
         'prompt': (
-            "The custom name for your AWS Accounts Connection Table. You must provide either an "
+            "The name for your AWS Accounts Connection. You must provide either an "
             "access key and secret key pair, or a source role, destination role, external id, and accounts "
-            "connection name."
+            "connection table name."
         ),
         'placeholder': "AWS_ACCOUNTS_DEFAULT_CONNECTION (NEEDED WITH SOURCE ROLE ARN, DESTINATION ROLE ARN, AND EXTERNAL ID)",
     }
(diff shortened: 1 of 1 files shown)


src/connectors/aws_inventory.py
```diff
@@ -95,14 +95,14 @@
     },
     {
         'type': 'str',
-        'name': 'accounts_identifier',
-        'title': "AWS Accounts Connection Identifier",
+        'name': 'accounts_connection_name',
+        'title': "AWS Accounts Table Name",
         'prompt': (
-            "The custom name for your AWS Accounts Connection, if you provided one. You must provide either an "
+            "The custom name for your AWS Accounts Connection Table. You must provide either an "
```
greg-snowflake: This should read "The name for your AWS Account Connection Table. [...]" instead of "The custom name"

src/connectors/aws_inventory.py
```diff
@@ -95,14 +95,14 @@
...
             "access key and secret key pair, or a source role, destination role, external id, and accounts "
-            "connection identifier."
+            "connection name."
```
greg-snowflake: should be "connection table name"

src/connectors/aws_inventory.py
```diff
@@ -192,22 +192,22 @@ def ingest(table_name, options):
     source_role_arn = options.get('source_role_arn')
     destination_role_name = options.get('destination_role_name')
     external_id = options.get('external_id')
-    accounts_identifier = options.get('accounts_identifier')
+    accounts_connection_name = options.get('accounts_connection_name')
```
greg-snowflake: After this line, can you add a quick check to see if the option starts with `data.`, and prepend `data.` if it doesn't? That way, it'll support the user either providing or omitting the schema, and will help prevent bug reports down the line.

src/connectors/aws_inventory.py
```diff
@@ -192,22 +192,26 @@ def ingest(table_name, options):
...
+
+    #Want to prepend 'data.' if it doesn't exist in the input
+    if accounts_connection_name[0:5] != 'data.':
```
greg-snowflake: quick note; this will result in the input `DATA.AWS_ACCOUNTS_DEFAULT_CONNECTION` turning into `data.DATA.AWS_ACCOUNTS_DEFAULT_CONNECTION`; you'll want to make sure you know what case the string is (which can be done by calling .lower() on it, I think) before you try to do a comparison.

also, you can use `accounts_connection_name.startswith('data.')` as an alternative to slicing the string, if you like.

PR #318, opened 2019-09-12
Rename the Cloudtrail comment to aws_cloudtrail

edulop91:
ModuleNotFoundError: No module named 'connectors.cloudtrail'


greg-snowflake:
Ah, good catch!

Issue #319, opened 2019-09-12
ec2_dispatch: UnboundLocalError: local variable 'results' referenced before assignment


edulop91:
```
[64] Error loading logs into AWS_ASSET_INV_EC2_DEFAULT_CONNECTION: 
[64] ./run.py:76 in <module>
    fire.Fire(main)
./../../venv/lib/python3.7/site-packages/fire/core.py:127 in Fire
    component_trace = _Fire(component, args, context, name)
./../../venv/lib/python3.7/site-packages/fire/core.py:366 in _Fire
    component, remaining_args)
./../../venv/lib/python3.7/site-packages/fire/core.py:542 in _CallCallable
    result = fn(*varargs, **kwargs)
./run.py:69 in main
    connectors_runner.main()
./../../venv/lib/python3.7/site-packages/runners/connectors_runner.py:75 in main
    Pool(DC_POOLSIZE).map(connection_run, tables)
./../../../../../usr/local/lib/python3.7/multiprocessing/context.py:119 in Pool
    context=self.get_context())
./../../../../../usr/local/lib/python3.7/multiprocessing/pool.py:176 in __init__
    self._repopulate_pool()
./../../../../../usr/local/lib/python3.7/multiprocessing/pool.py:241 in _repopulate_pool
    w.start()
./../../../../../usr/local/lib/python3.7/multiprocessing/process.py:112 in start
    self._popen = self._Popen(self)
./../../../../../usr/local/lib/python3.7/multiprocessing/context.py:277 in _Popen
    return Popen(process_obj)
./../../../../../usr/local/lib/python3.7/multiprocessing/popen_fork.py:20 in __init__
    self._launch(process_obj)
./../../../../../usr/local/lib/python3.7/multiprocessing/popen_fork.py:74 in _launch
    code = process_obj._bootstrap()
./../../../../../usr/local/lib/python3.7/multiprocessing/process.py:297 in _bootstrap
    self.run()
./../../../../../usr/local/lib/python3.7/multiprocessing/process.py:99 in run
    self._target(*self._args, **self._kwargs)
./../../../../../usr/local/lib/python3.7/multiprocessing/pool.py:121 in worker
    result = (True, func(*args, **kwds))
./../../../../../usr/local/lib/python3.7/multiprocessing/pool.py:44 in mapstar
    return list(map(*args))
./../../venv/lib/python3.7/site-packages/runners/connectors_runner.py:64 in connection_run
[...]

PR #320, opened 2019-09-12
Small fixes to the ec2 inventory connector

edulop91:
Fixes https://github.com/snowflakedb/SnowAlert/issues/319

Let me know if the auto-formatting is annoying and I can disable it!


greg-snowflake:
I saw the bug report in my email and wrote https://github.com/snowflakedb/SnowAlert/pull/321 before I realized you had a patch in for it as well! The only difference between my fix and yours is that mine keeps a running total of the results, instead of just returning the most recent one discovered; this will probably help avoid another bug report down the line.

PR #325, opened 2019-09-16
Upgrade snowflake-connector-python to 1.9.0

edulop91:
Snowflake support suggested I upgrade client version to latest. This seems to help with the JWT invalid token issue somewhat.


andrey-snowflake:
good call, ty

PR #326, opened 2019-09-16
Add pybrake to requirements.txt

edulop91:
Seems like I broke the build, this should fix it.


andrey-snowflake:
the fix was something else, but this is good to include, as well :-)

PR #328, opened 2019-09-17
Create ec2-missing-osquery.sql

omer-snowflake:
Adding violation rule for visibility gap detecting server in AWS missing osquery


rules/visibility/ec2-missing-osquery.sql
```diff
@@ -0,0 +1,53 @@
...
+FROM (
+  SELECT timestamp as snapshot_at
+       , instance:InstanceId::String as instances
+       , instance:Tags::String as tags
+       , instance:State:Name::String as state
+       , instance:"AccountId" as accountid
+       , instance:Platform::String as platform
+  , instance:"LaunchTime" as launch_time
+       , instance as details
+  FROM snowalert.base_data.ec2_instance_snapshots_t as snap
```
edulop91: The only thing I ran into was this `base_data.ec2_instance_snapshots_t` and `security.prod.osquery_v` seem to deviate from the current naming conventions.` I think some columns might be different because of this.

PR #332, opened 2019-09-19
v1.8.7 Improved Connectors, Handlers, Runners

andrey-snowflake:
# Data Connectors

695975f Adds Fire library to the runner, to run connections separately from CLI
1e8ebbf Updates Okta Connector to include deprovisioned users
977061f Fixed AWS CloudTrail errors when mfaAuthenticated is str (#336)
125a848 Fixes AWS Flow UI Role field name (#330)
68b3b68 Adds IAM Connection Type to AWS Inventory (#337)
21d2593 Adds mypy checking to connectors modules
834610d Adds gov cloud support to Azure ingestion
be43920 Adds account IDs to AWS Inventory EC2 and ELB ingestion (#344)
b2f26f9 Updates Azure client dependencies

# Handlers

c64ccf2 Improves Jira handler's custom fields options
8a278ea Adds and fixes SMTP handler

# Query Runners

9f03e11 Adds custom Alert cutoff time via env var
e41ea84 Fixes Violation runner in single violation run mode

# Misc

8f3d5e6 Standardizes Python code formatting with [Black](https://pypi.org/project/black/)

# Thanks

Thank you @alldoami, @edulop91, and @sf-bhushanchitte for contributions to this release!


src/connectors/aws_inventory.py
```diff
@@ -144,7 +146,20 @@
...
+    'IAM': [
+        ('raw', 'VARCHAR'),
+        ('ingested_at', 'TIMESTAMP_LTZ'),
+        ('path', 'VARCHAR'),
+        ('user_name', 'VARCHAR'),
+        ('user_id', 'VARCHAR'),
+        ('arn', 'VARCHAR'),
+        ('create_date', 'TIMESTAMP_LTZ'),
+        ('password_last_used', 'TIMESTAMP_LTZ'),
+        ('account_id', 'STRING(32)'),
```
greg-snowflake: Why is the length for IAM account_id 32, when it's 30 in the other tables?

andrey-snowflake: not sure. @edulop91 should we update the others, as well?

edulop91: Power of 2 seemed nice, not sure if it has any real implications though. Should I make all of them 30?

andrey-snowflake: Last we discussed Greg likes powers of 2 and I think they're overrated, but don't think any of us have a strong preference :-)

greg-snowflake: I'm mostly of the opinion that consistency is better than truth - I don't have strong opinions on how long the column should be once we get past basic issues like "is it long enough" but i firmly believe that if two columns are meant to hold the same thing they should be the same size, whatever that size is.

PR #333, opened 2019-09-19
Make cutoff time configurable and add support for a JIRA custom field

sf-bhushanchitte:
- Making grouping cutoff time configurable via environment variable while keeping backwards compatible
- Adding support for JIRA custom field "Area"

### DIFF
--- a/src/runners/handlers/jira.py
+++ b/src/runners/handlers/jira.py
@@ -115,10 +115,11 @@ def create_jira_ticket(alert, assignee=None, custom_field=None):
     env_fields = os.environ.get('SA_JIRA_CUSTOM_FIELDS')
     if env_fields or custom_field:
         custom_fields = [f.split('=') for f in env_fields.split(';')]
-        if custom_field:
-            issue_params[f'customfield_{custom_field["id"]}'] = {'value': custom_field['value']}
         for field_id, field_value in custom_fields:
             issue_params[f'customfield_{field_id}'] = {'value': field_value}
+
+        if custom_field:
+            issue_params[f'customfield_{custom_field["id"]}'] = {'value': custom_field['value']}
 
     new_issue = jira.create_issue(**issue_params)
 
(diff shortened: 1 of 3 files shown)


andrey-snowflake:
Thanks! I noticed that the way you are setting custom fields is a little different (better) than the existing code for the custom_field argument defined in the param. Could you please either add automated tests or let me know how you tested it?

I've made your code a little more general (removing the field ID is specific to Snowflake), and have integrated the custom fields defined in ENV (but the SA administrator) with the custom field(s) passed in via the `custom_field` rule parameter.

Let me know if this makes sense and whether you need help testing, please!

src/runners/handlers/jira.py
```diff
@@ -104,18 +104,25 @@ def create_jira_ticket(alert, assignee=None, custom_field=None):
...
+
+    issue_params = {
+        'project': PROJECT,
+        'issuetype': {'name': 'Story'},
+        'summary': alert['TITLE'],
+        'description': body,
+    }
+
+    env_fields = os.environ.get('SA_JIRA_CUSTOM_FIELDS')
+    if env_fields or custom_field:
```
sf-bhushanchitte: Thanks for this change. I thought a very similar approach to make my changes more general, but then backed off thinking there may not be a need to support for those many custom field.

src/runners/handlers/jira.py
```diff
@@ -104,18 +104,25 @@ def create_jira_ticket(alert, assignee=None, custom_field=None):
...
+        custom_fields = [f.split('=') for f in env_fields.split(';')]
+        if custom_field:
+            issue_params[f'customfield_{custom_field["id"]}'] = {'value': custom_field['value']}
+        for field_id, field_value in custom_fields:
+            issue_params[f'customfield_{field_id}'] = {'value': field_value}
+        issue_params['customfield_11401'] = {'value': os.environ.get('CUSTOM_FIELD_AREA')}
```
sf-bhushanchitte: I guess we need to remove this line as above generic logic will take care of all the custom fields?

andrey-snowflake: yup, just wanted to split it into separate commit :-)

src/runners/handlers/jira.py
andrey-snowflake: We may not have more than one, but we do want to keep it extensible to Jira's outside of our main corp one.

PR #337, opened 2019-09-23
[DC] IAM Connector

alldoami:
Tested with python:
```
import src.connectors.aws_inventory as aws_inventory
options = {"connection_type": "IAM", "aws_access_key": "xxxx", "aws_secret_key": "xxxx", "accounts_connection_name": "AWS_ACCOUNTS_DEFAULT_CONNECTION"}
print([x for x in aws_inventory.ingest("AWS_ASSET_INV_IAM_DEFAULT_CONNECTION", options)])
```


src/connectors/aws_inventory.py
```diff
@@ -327,6 +373,38 @@ def elb_dispatch(landing_table, aws_access_key='', aws_secret_key='', accounts=N
...
+    users = get_iam_users(
+        aws_access_key=aws_access_key,
+        aws_secret_key=aws_secret_key,
+        session=session,
+        account=account
+    )
+
+    monitor_time = datetime.utcnow().isoformat()
+
+    print(db.derive_insert_select(LANDING_TABLES_COLUMNS))
```
edulop91: rm these print statements

src/connectors/aws_inventory.py
```diff
@@ -327,6 +373,38 @@ def elb_dispatch(landing_table, aws_access_key='', aws_secret_key='', accounts=N
...
+            monitor_time,
+            row['Path'],
+            row['UserName'],
+            row['UserId'],
+            row.get('Arn'),
+            row['CreateDate'],
+            row.get('PasswordLastUsed'))
+            for row in users
+        ],
+        select=db.derive_insert_select(LANDING_TABLES_COLUMNS),
```
edulop91: `LANDING_TABLE_COLUMNS` references all the landing tables for all types of ingestion. I think you want `LANDING_TABLES_COLUMNS['IAM']`

src/connectors/aws_inventory.py
```diff
@@ -405,6 +483,33 @@ def ingest_elb(landing_table, aws_access_key=None, aws_secret_key=None, session=
     )
     return len(elbs)
 
+def get_iam_users(aws_access_key=None, aws_secret_key=None, session=None, account=None):
+    client = boto3.client(
```
edulop91: seems like this client is unused, you can remove it

src/connectors/aws_inventory.py
```diff
@@ -145,6 +146,17 @@
         ('region_name', 'STRING(16)'),
         ('scheme', 'STRING(30)'),
         ('vpc_id', 'STRING(30)')
+    ],
+    # IAM Users
+    'IAM': [
+        ('raw', 'VARCHAR'),
+        ('ingested_at', 'TIMESTAMP_LTZ'),
+        ('Path', 'VARCHAR'),
```
edulop91: PATH

src/connectors/aws_inventory.py
```diff
@@ -145,6 +146,17 @@
...
+        ('UserName', 'VARCHAR'),
```
edulop91: USER_NAME

src/connectors/aws_inventory.py
```diff
@@ -145,6 +146,17 @@
...
+        ('UserId', 'VARCHAR'),
```
edulop91: USER_ID

src/connectors/aws_inventory.py
```diff
@@ -145,6 +146,17 @@
...
+        ('Arn', 'VARCHAR'),
```
edulop91:
```suggestion
        ('arn', 'VARCHAR'),
```

src/connectors/aws_inventory.py
```diff
@@ -145,6 +146,17 @@
...
+        ('CreateDate', 'TIMESTAMP_LTZ'),
```
edulop91:
```suggestion
        ('create_date', 'TIMESTAMP_LTZ'),
```

src/connectors/aws_inventory.py
```diff
@@ -145,6 +146,17 @@
...
+        ('PasswordLastUsed', 'TIMESTAMP_LTZ')
```
edulop91:
```suggestion
        ('password_last_used', 'TIMESTAMP_LTZ')
```

src/connectors/aws_inventory.py
edulop91: all of these for convention and consistency

src/connectors/aws_inventory.py
alldoami: We still need to initialize it before the if statement right?

greg-snowflake:
This looks good to me; can you add a note describing how you tested it?

edulop91:
python:
```
import src.connectors.aws_inventory as aws_inventory
options = {"connection_type": "IAM", "aws_access_key": "xxxx", "aws_secret_key": "xxxx", "accounts_connection_name": "AWS_ACCOUNTS_DEFAULT_CONNECTION"}
print([x for x in aws_inventory.ingest("AWS_ASSET_INV_IAM_DEFAULT_CONNECTION", options)])
```

andrey-snowflake:
lgtm, although I think this is doubling up with our AWS Config ingestion, which contains IAM changes, as well. there's also a config snapshot API call we could use, but with other APIs this has been the most thorough approach.

since there isn't anything super urgent on the release line right now, I think we can include this in v1.8.7 and push back the release a week (leaving next week for testing)?

omer-snowflake:
@andrey-snowflake actually this is information that is not brought back by Config. As I was working on CIS AWS Benchmark last week I saw we were missing important details like MFA and last login for users. This was a story in our upcoming sprint so we'll definitely want this one in the next release (pushing it back a week sounds fine to allow for testing). Thank you @alldoami and @edulop91 for putting this one together! I'll share the CIS compliance rules when we have them working with this data.

src/connectors/aws_inventory.py
```diff
@@ -405,6 +483,30 @@ def ingest_elb(landing_table, aws_access_key=None, aws_secret_key=None, session=
...
+    page_iterator = paginator.paginate()
+    results = [
+        user
+        for page in page_iterator
+        for user in page['Users']
+    ]
+
+    for user in results:
+        if account:
+            results['Account'] = account
```
edulop91:
```suggestion
            user['Account'] = account
```

PR #338, opened 2019-09-27
[DC] AWS AMI Inventory

edulop91:
launched the webui locally to create the connector table.

then ran the following from the python interpreter to check it works:
```
import src.connectors.aws_inventory as aws_inventory
options = {"connection_type": "AMI", "aws_access_key": "XXXXX", "aws_secret_key": "XXXX", "accounts_connection_name": "AWS_ACCOUNTS_DEFAULT_CONNECTION"}
print([x for x in aws_inventory.ingest("AWS_ASSET_INV_AMI_DEFAULT_CONNECTION", options)])
```


edulop91:
let me know if the auto formatting is annoying and i'll look into fixing it

andrey-snowflake:
Let's not worry about it and add a PEP-8 post-commit format script if it gets annoying.

The code looks fine but since we don't have a need for some of these, going to rely heavily on your testing here.

edulop91:
sounds good!
I've been running it manually from my command line with an AWS keypair. Waiting until we merge this to run it as an assume role -- I can then fix whatever issues come up if any do.

PR #343, opened 2019-10-02
[DC] AWS AMI Inventory

edulop91:

PR #346, opened 2019-10-07
[DC] AWS AMI Inventory

andrey-snowflake:
rebased version of #343


src/connectors/aws_inventory.py
```diff
@@ -363,6 +389,62 @@ def ec2_dispatch(
...
+            except Exception as e:
+                db.insert(
+                    AWS_ACCOUNTS_METADATA,
+                    values=[(datetime.utcnow(), RUN_ID, id, name, 0, e)],
+                    columns=[
+                        'snapshot_at',
+                        'run_id',
+                        'account_id',
+                        'account_alias',
+                        'ec2_count',
```
greg-snowflake: I don't think you want ec2_count here.

PR #365, opened 2019-12-13
fix:dc:aws_collect: Cartesian product between methods and accounts works correctly now

edulop91:
For some reason, the previous code would only the combination of the first method for all accounts. 
This fix should take `methods X accounts`.


edulop91:
can't repro, going to close for now

PR #366, opened 2019-12-18
[fix] allowed command line input to alert query runner

kuannie1:


andrey-snowflake:
Sorry I missed this -- migrating to a direct runner rather than going through `run.py` isn't urgent, but could be a welcome improvement. Happy to chat if you'd like to figure out exactly how we'd want to migrate.

Issue #360, opened 2019-11-25
Data Connector for Salesforce Event Monitoring


hh-jamesweakley:
See training material here: https://trailhead.salesforce.com/en/content/learn/modules/event_monitoring

These are the events that come with the Shield add-on, that act as an audit trail of user activity. The use case would be monitoring for suspicious CRM activity (e.g. viewing a suspiciously large number of different customer records)

User Login History (https://help.salesforce.com/articleView?id=users_login_history.htm&type=5) is another possible data source but is not as rich.

Note that someone from my company is likely to implement request this in the coming months, but registering interest in the meantime.

omer-snowflake:
Thanks James, we have Shield too but haven't integrated it yet so this would be great to have. Is the person at your company going to implement to integration as a data connector to Snowflake?

hh-jamesweakley:
Sorry @omer-snowflake, missed this comment at the time. Yes, we're planning to build a SnowAlert data connector for Salesforce Event Monitoring.

andrey-snowflake:
Thanks again for the contribution, @hh-jamesweakley! Feel free to resolve when you have a working deployment with released code or we can when we do.

andrey-snowflake:
Could you add a doc in `./docs/connectors` or write up here some instructions w.r.t. recommended usage for the connector? Should we be creating a service user to use this, and then the service user creates an API key? 

![image](https://user-images.githubusercontent.com/42752788/71841484-59411b00-3074-11ea-8a82-b475fb5fb6c4.png)

andrey-snowflake:
Looking good, thanks again!

PR #370, opened 2020-01-06
Salesforce Event Log data connector

hh-jamesweakley:
This PR adds a new data connector for collecting hourly Salesforce Event Log data via API. Commit is based off the v1.9.2 branch which I assume will be merged to master next.

All ~50 different [event types](https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_eventlogfile_supportedeventtypes.htm) are loaded into a single table by converting each downloaded log from CSV to JSON prior to upload.

No alerts are implemented yet, these will be added later after sufficient production data is collected.

A few notes:
- A table stage is used instead of creating a named stage
- CONTRIBUTING.md mentions implementing a 'test' function to test the connection, but it isn't supplied with the connection options so wasn't sure how this was possible. I considered doing a Salesforce connection test in the 'connect' function prior to table creation, but the options passed in there are pre-encrypted. So I just left the connection test out, but happy to take a recommendation on how to approach this.
- It would also be fairly simple to extract Salesforce configuration (e.g. AuthConfig objects). I figured that would be a different data connector, but let me know if you prefer it all in one.

Issue #360


hh-jamesweakley:
image

image

PR #385, opened 2020-02-14
[feature] add AWS Inspector to AWS Collector Functionality

kuannie1:


src/connectors/aws_collect.py
```diff
@@ -1008,6 +1069,10 @@ def process_aws_response(task, page):
             ):
                 yield x
 
+    except ParamValidationError as e:
+        # hack(aku): Should do something about e
+        pass
+
```
kuannie1: @andrey-snowflake This error came up while I was trying to insert list-findings results into describe-findings (sometimes I would get an empty list). This exception doesn't have a `response` attribute.

migrations/v1_9_1-v1_9_2.md
```diff
@@ -45,5 +45,39 @@ CREATE TABLE data.aws_collect_iam_get_role_policy (
 ;
 GRANT SELECT, INSERT ON data.aws_collect_iam_get_role_policy TO ROLE snowalert;
 
+CREATE TABLE data.aws_collect_inspector_list_findings (
```
andrey-snowflake: move into `migrations/v1_9_2-v1_9_3.md`, please

src/connectors/aws_collect.py
```diff
@@ -454,6 +455,34 @@
         ('data_resources', 'VARIANT'),
         ('exclude_management_event_sources', 'VARIANT'),
     ],
+    # https://docs.aws.amazon.com/cli/latest/reference/inspector/list-findings.html
+    'inspector_list_findings': [
+        ('finding_arns', 'VARIANT')
```
andrey-snowflake: I think these all require `recorded_at`, `account_id`, and `region` columns, as the others. You can copy-paste them from above.

PR #390, opened 2020-03-10
Azure CIS 2.1 to 3.8

andrey-snowflake:
Tested by running queries against our collected data, double checking nothing seemed too surprising.

### DIFF
--- a/packs/cis_azure_pack.sql
+++ b/packs/cis_azure_pack.sql
@@ -4,7 +4,7 @@ CREATE OR REPLACE VIEW rules.AZURE_CIS_1_3_VIOLATION_QUERY COPY GRANTS
   @tags cis, azure, iam'
 AS
 SELECT 'PLSEUOLMOH' AS query_id
-     , 'Azure CIS 1.3: Ensure that there are no guest users' AS title
+     , 'Azure CIS 1.3: that there are no guest users' AS title
      , OBJECT_CONSTRUCT(
          'cloud', 'azure',
          'account', tenant_id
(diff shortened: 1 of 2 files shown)


packs/cis_azure_pack.sql
```diff
@@ -74,3 +74,192 @@ WHERE 1=1
...
+    subscription_id,
+    name,
+    type,
+    sku,
+    properties:parameters params
+  FROM data.azure_collect_policy_assignments
+  WHERE recorded_at > CURRENT_DATE - 2
+)
+WHERE 1=1
+  AND 'Disabled' IN (
```
mf-sf: I had a look at the raw data and I think that most of the times (except tenant:d479c7c9-2632-445a-b22d-7c19e68774f6 subscription:f4b00c5f-f6bf-41d6-806b-e1cac4f1f36f) the parameters fields is an empty dictionary, I'm not sure what it means, my guess is that this feature is disabled in the subscription. I think we would have to check with the Azure team to confirm. If that's the case I think we should trigger this on empty dictionary AND incorrect value in the dictionary field.

Probably this needs to be split into one violation per CIS entry to show up correctly in the dashboards :(

packs/cis_azure_pack.sql
```diff
@@ -74,3 +74,192 @@ WHERE 1=1
...
+    params:vulnerabilityAssesmentMonitoringEffect,
+    params:storageEncryptionMonitoringEffect,
+    params:jitNetworkAccessMonitoringEffect,
+    params:adaptiveApplicationControlsMonitoringEffect,
+    params:sqlAuditingMonitoringEffect,
+    params:sqlEncryptionMonitoringEffect
+ )
+;
+
+CREATE OR REPLACE VIEW rules.AZURE_CIS_2_16_TO_19_VIOLATION_QUERY COPY GRANTS
```
mf-sf: Also for this I believe we should split into one violation per CIS benchmark

packs/cis_azure_pack.sql
andrey-snowflake: Good catch -- the CIS doc says "Ensure the output of the below command is not Disabled or Empty", so going to go with that rather than reaching out to their support, unless you'd prefer to double check with them?

mf-sf: nope, go for it!

PR #392, opened 2020-03-14
OG-001: add Pager Duty handler

olegg:


olegg:
@andrey-snowflake - ^^ :)

andrey-snowflake:
Yikes -- just realized I should have merged into v1.9.4 branch. What happens now is that v1.9.3 will be re-build to include this, but I'll still add / credit this in the v1.9.4 release notes.

PR #400, opened 2020-05-04
add logic to re-run unsuccessful handlers

mikeurbanski1:
I've added logic to retry handlers where `success = false`.

Before this, an error in a handler (in my case, errors not correctly handled by the slack handler, which I will submit a PR for shortly) would result in an entry in `alerts.HANDLED`, which would prevent that alert from ever being handled again.

The solution would be to set `alerts.HANDLED` to `NULL`, but that's not a great option when you have multiple handlers, and the other handlers succeeded. It's also not preferable to have that manual step in general.

This change extends the query to return every unhandled alert, and also any alert with a failed handler. In the latter case, it will skip successful handlers and only retry the failed ones.

The result is that a fix for the error (for example, fixing that slack error where the json string was not escaped properly, and single quotes broke the query, or fixing some error in the UDF itself) will result in just the failed the handler being fired on the next run.

### DIFF
--- a/src/runners/alert_dispatcher.py
+++ b/src/runners/alert_dispatcher.py
@@ -8,8 +8,10 @@ from .utils import apply_some, json_dumps
 
 GET_ALERTS_QUERY = f"""
 SELECT *
-FROM results.alerts
-WHERE IFF(alert:HANDLERS IS NULL, ticket IS NULL, handled IS NULL)
+FROM results.alerts a1
+WHERE IFF(alert:HANDLERS IS NULL, ticket IS NULL, handled IS NULL or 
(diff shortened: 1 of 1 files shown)


sfc-gh-afedorov:
thanks -- this seems like a solid improvement to our "v0" version. I can merge it if you add tests, or I can add some when I get the chance.

sfc-gh-afedorov:
we chatted more about this and would rather not re-run handlers automatically without more consideration as to how to signal failure cases. internally, we've got alerts to notify us other alerts having failed handlers, as this is usually a case where human intervention is needed, and the infra operators can reset handler columns manually to attempt to re-deliver once that is fixed.

PR #402, opened 2020-05-04
add explicit 'from time' argument; defend from SQL injection

mikeurbanski1:
This change adds a few small features:

1. Explicit support for an `SA_ALERT_FROM_TIME` environment variable when the user would rather specify a given range instead of an end time and cutoff period
2. Proper handling of arguments to prevent SQL injection
3. Locking the "to" time to a consistent timestamp fur the duration of the run

**1. Support for "from time"**

An example use case is for automated periodic runs while ensuring an error does not result in a missed window. For example, say the normal execution of the task is hourly and looks at alerts from the last hour. Upon successful completion of the task, it can update the `alert_from_time` to be the previous run's `alert_to_time`. If the task fails, then the old `alert_from_time` will remain, and the next run of the task will include the "missed" window.

**2. SQL injection prevention**

Arguments are validated using formatting rules. The cutoff minutes value is converted to an int; timestamps are checked using `strftime`. If any of these validations fails, then the arguments are invalid. This also makes the job fail before running any alert queries, making errors in the input values easier to detect.

**3. Consistent timestamp**

Instead of inserting `CURRENT_TIMESTAMP` into a query if no "to time" is given, this change gets the current timestamp at the start of the run, and then inserts it into each query. This ensures each alert query runs over the exact same window, avoiding the case where a long-running alert query can throw off the time window of subsequent alert queries.

Note that this logic is now convoluted enough that it should probably be refactored into a helper or utility method. I do have other code that reuses these global variables once they are set, and I am not 100% sure that there is no other logic that relies on this code running once, so I left the structure as-is.

### DIFF
--- a/src/runners/alert_queries_runner.py
+++ b/src/runners/alert_queries_runner.py
@@ -28,7 +28,7 @@ from runners.helpers import db, log
 # - SA_ALERT_CUTOFF_MINUTES=x: use the last x minutes.
 # - SA_ALERT_FROM_TIME=x: use the window from x to now.
 # - SA_ALERT_TO_TIME=x: use the 90 minutes before / up to x.
-# - SA_ALERT_TO_TIME=x, SA_ALERT_CUTOFF_MINUTES=y: use the y minutes before / up to x.
+# - SA_ALERT_CUTOFF_MINUTES=x, SA_ALERT_TO_TIME=y: use the x minutes before / up to y.
 # - SA_ALERT_FROM_TIME=x, SA_ALERT_TO_TIME=y: use the window x to y.
 
 # We'll also do some simple checks to prevent against SQL injections, since these values will be inserted as-is into
--- a/src/runners/alert_queries_runner.py
+++ b/src/runners/alert_queries_runner.py
@@ -2,7 +2,6 @@
 
 import datetime
 import os
-import pytz
 from multiprocessing import Pool
 from typing import Any, Dict
 
--- a/src/runners/alert_queries_runner.py
+++ b/src/runners/alert_queries_runner.py
@@ -16,53 +16,21 @@ from runners.config import (
 from runners.helpers import db, log
 
 
-# There are three time window properties that will be combined in different ways depending on which are present.
-# - SA_ALERT_FROM_TIME: a timestamp specifying the start of the window. If omitted, use SA_ALERT_CUTOFF_MINUTES relative
-# to SA_ALERT_TO_TIME.
-# - SA_ALERT_TO_TIME: a timestamp specifying the end of the window. If omitted, use the current timestamp.
-# - SA_ALERT_CUTOFF_MINUTES: the length of the window relative to SA_ALERT_TO_TIME; only used if SA_ALERT_FROM_TIME is
-# not present. Defaults to 90.
-#
-# So, the possible combinations to specify a window are:
-# - All options omitted: use the last 90 minutes.
-# - SA_ALERT_CUTOFF_MINUTES=x: use the last x minutes.
-# - SA_ALERT_FROM_TIME=x: use the window from x to now.
-# - SA_ALERT_TO_TIME=x: use the 90 minutes before / up to x.
-# - SA_ALERT_CUTOFF_MINUTES=x, SA_ALERT_TO_TIME=y: use the x minutes before / up to y.
-# - SA_ALERT_FROM_TIME=x, SA_ALERT_TO_TIME=y: use the window x to y.
+# Three envars, two ways to set window properties:
+# 1. by start & end: (SA_ALERT_FROM_TIME, SA_ALERT_TO_TIME)
+# 2. by end & duration: (SA_ALERT_TO_TIME, SA_ALERT_CUTOFF_MINUTES)
 
-# We'll also do some simple checks to prevent against SQL injections, since these values will be inserted as-is into
-# the query.
-
-# TODO this should probably be moved to a method, but there may be other logic that depends on these global variables
-# or the fact that this would only be executed once on import in its current location (as opposed to calling an
-# initialization method from the main method which could end up getting repeated, for example).
+# SA_ALERT_TO_TIME default is current time
+# SA_ALERT_CUTOFF_MINUTES default is 90
 
 ALERT_CUTOFF_MINUTES = int(os.environ.get('SA_ALERT_CUTOFF_MINUTES', -90))
 if ALERT_CUTOFF_MINUTES > 0:
     ALERT_CUTOFF_MINUTES = -ALERT_CUTOFF_MINUTES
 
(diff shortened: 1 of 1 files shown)


mikeurbanski1:
One issue I've had periodically since I made this change was that running the alert queries hangs, gives an error about the JWT token, and then continues normally:

<details>
<summary>Click for logs</summary>

```
[7] Alerts from time: DATEADD(minute, -2880, '2020-05-05 07:00:30.263 -0700')
[7] Alerts to time: '2020-05-05 07:00:30.263 -0700'
[7] STARTING RUN WITH ID 242e1e5f06794bc0a9490532dc208702
[7] got command alerts
[7] Loaded 2 views, 2 were 'ALERT_QUERY' rules.
[20] GDRIVE_DOCUMENT_SHARED_TO_SELF_ALERT_QUERY processing...
[19] GDRIVE_DOCUMENT_SHARED_TO_PRIVATE_EMAIL_ALERT_QUERY processing...
[20] GDRIVE_DOCUMENT_SHARED_TO_SELF_ALERT_QUERY created 1, updated 0 rows.
[20] GDRIVE_DOCUMENT_SHARED_TO_SELF_ALERT_QUERY metadata recorded.
[20] GDRIVE_DOCUMENT_SHARED_TO_SELF_ALERT_QUERY done.
[19] ./run.py:79 in <module>
    fire.Fire(main)
./../../venv/lib/python3.7/site-packages/fire/core.py:127 in Fire
    component_trace = _Fire(component, args, context, name)
./../../venv/lib/python3.7/site-packages/fire/core.py:366 in _Fire
    component, remaining_args)
./../../venv/lib/python3.7/site-packages/fire/core.py:542 in _CallCallable
    result = fn(*varargs, **kwargs)
./run.py:58 in main
    alert_queries_runner.main()
./../../venv/lib/python3.7/site-packages/runners/alert_queries_runner.py:188 in main
    metadata = Pool(POOLSIZE).map(create_alerts, rules)
./../../../../../usr/local/lib/python3.7/multiprocessing/context.py:119 in Pool
    context=self.get_context())
./../../../../../usr/local/lib/python3.7/multiprocessing/pool.py:176 in __init__
    self._repopulate_pool()
./../../../../../usr/local/lib/python3.7/multiprocessing/pool.py:241 in _repopulate_pool
    w.start()
./../../../../../usr/local/lib/python3.7/multiprocessing/process.py:112 in start
    self._popen = self._Popen(self)
./../../../../../usr/local/lib/python3.7/multiprocessing/context.py:277 in _Popen
    return Popen(process_obj)
./../../../../../usr/local/lib/python3.7/multiprocessing/popen_fork.py:20 in __init__
    self._launch(process_obj)
./../../../../../usr/local/lib/python3.7/multiprocessing/popen_fork.py:74 in _launch
    code = process_obj._bootstrap()
./../../../../../usr/local/lib/python3.7/multiprocessing/process.py:297 in _bootstrap
    self.run()
./../../../../../usr/local/lib/python3.7/multiprocessing/process.py:99 in run
    self._target(*self._args, **self._kwargs)
./../../../../../usr/local/lib/python3.7/multiprocessing/pool.py:121 in worker
    result = (True, func(*args, **kwds))
./../../../../../usr/local/lib/python3.7/multiprocessing/pool.py:44 in mapstar
    return list(map(*args))
./../../venv/lib/python3.7/site-packages/runners/alert_queries_runner.py:162 in create_alerts
    insert_count, update_count = merge_alerts(rule_name, ALERTS_FROM_TIME)
./../../venv/lib/python3.7/site-packages/runners/alert_queries_runner.py:138 in merge_alerts
    result = db.execute(sql, fix_errors=False).fetchall()
./../../venv/lib/python3.7/site-packages/runners/helpers/db.py:161 in execute
    ctx, query = connect(), ctx
./../../venv/lib/python3.7/site-packages/runners/helpers/db.py:122 in connect
    connection = retry(connect)
./../../venv/lib/python3.7/site-packages/runners/helpers/db.py:55 in retry
    log.error(e)
--- printed exception w/ trace ---
./../../venv/lib/python3.7/site-packages/runners/helpers/db.py:46 in retry
    return f()
./../../venv/lib/python3.7/site-packages/runners/helpers/db.py:118 in connect
    network_timeout=TIMEOUT,
./../../venv/lib/python3.7/site-packages/snowflake/connector/__init__.py:33 in Connect
    return SnowflakeConnection(**kwargs)
./../../venv/lib/python3.7/site-packages/snowflake/connector/connection.py:187 in __init__
    self.connect(**kwargs)
./../../venv/lib/python3.7/site-packages/snowflake/connector/connection.py:484 in connect
    self.__open_connection()
./../../venv/lib/python3.7/site-packages/snowflake/connector/connection.py:700 in __open_connection
    self.__authenticate(auth_instance)
./../../venv/lib/python3.7/site-packages/snowflake/connector/connection.py:931 in __authenticate
    session_parameters=self._session_parameters,
./../../venv/lib/python3.7/site-packages/snowflake/connector/auth.py:308 in authenticate
    u'sqlstate': SQLSTATE_CONNECTION_WAS_NOT_ESTABLISHED,
./../../venv/lib/python3.7/site-packages/snowflake/connector/errors.py:100 in errorhandler_wrapper
    connection.errorhandler(connection, cursor, errorclass, errorvalue)
./../../venv/lib/python3.7/site-packages/snowflake/connector/errors.py:73 in default_errorhandler
    done_format_msg=errorvalue.get(u'done_format_msg'))
snowflake.connector.errors.DatabaseError: 250001 (08001): Failed to connect to DB: gl56520.us-east-1.snowflakecomputing.com:443. JWT token is invalid.
[19] GDRIVE_DOCUMENT_SHARED_TO_PRIVATE_EMAIL_ALERT_QUERY created 1, updated 3 rows.
[19] GDRIVE_DOCUMENT_SHARED_TO_PRIVATE_EMAIL_ALERT_QUERY metadata recorded.
[19] GDRIVE_DOCUMENT_SHARED_TO_PRIVATE_EMAIL_ALERT_QUERY done.
[7] RUN metadata recorded.
```
</details>

So, it runs the first alert query, which returns results that are processed normally. Then it hangs for several seconds, and this error occurs. Then it seems to reconnect and continue normally. 

I'm fairly certain that no query is actually failing during this time; I think the connection is just denying and then getting recreated. It's possible that one of the intermediate queries that doesn't actually get logged is failing.

I am not really sure why adding the query to get the current DB-specific timestamp would cause this issue, but I've only ever encountered this since making this change. However, it also doesn't happen every time, so it's hard to troubleshoot.

If you have any ideas for why this change would cause this issue, let me know and I will make the update. But, I do not have any ideas at this point.

sfc-gh-afedorov:
We've seen the JWT token errors before, and have a few ideas how to try mitigating them, but are also considering them a downstream error that might get fixed in the Python connector. This might be more urgent than we realize, but as of now I think it's mitigated via db retries.

PR #415, opened 2020-07-15
Add alert timestamps and ids

sfc-gh-afedorov:


sfc-gh-afedorov:
deprecating Python after migration to Task-based runners

PR #419, opened 2020-07-27
[handlers] Refactor Slack handler

sfc-gh-afedorov:


src/runners/handlers/slack.py
```diff
@@ -8,37 +8,28 @@
 from runners.helpers import vault
 
 
-def message_template(vars):
+def message_template(alert, properties, user, template):
     payload = None
 
     # remove handlers data, it might contain JSON incompatible strucutres
-    vars['alert'].pop('HANDLERS')
+    alert.pop('HANDLERS')
```
jmak123, line 15: I think this line might fail when multiple Slack handlers are present in the alert. The first call will pop ['HANDLERS'] but the second call so on will have nothing to pop and return error

sfc-gh-afedorov: good catch -- not sure if this would cause a bug just yet, but mutating the alert this way is definitely not part of the handler contract, so might cause issues down the line.

sfc-gh-afedorov:
this will need to be smaller refactors / made bug #439

PR #420, opened 2020-07-28
[dc] Adds Duo admin inventory connector

sfc-gh-mpolyakova:
tested by running in dev, got command —

	CREATE TABLE data.duo_default_admins_connection(raw VARIANT, recorded_at TIMESTAMP_LTZ DEFAULT CURRENT_TIMESTAMP)
	COMMENT='---
	duo_integration_key: asdfasdf
	duo_key: asdfasdf
	module: duo
	name: default
	subdomain: test
	';

and ran these to test rest —

	INSERT INTO data.duo_default_admins_connection (raw) SELECT PARSE_JSON(column1) FROM VALUES ('{}');

	SELECT * FROM data.duo_default_admins_connection;

got result as expected

### DIFF
--- a/src/connectors/duo_collect.py
+++ b/src/connectors/duo_collect.py
@@ -42,7 +42,10 @@ CONNECTION_OPTIONS = [
     },
 ]
 
-LANDING_ADMIN_TABLE_COLUMNS = [('raw', 'VARIANT', 'RECORDED_AT', 'TIMESTAMP_LTZ')]
+LANDING_ADMIN_TABLE_COLUMNS = [
+    ('raw', 'VARIANT'),
+    ('recorded_at', 'TIMESTAMP_LTZ DEFAULT CURRENT_TIMESTAMP'),
+]
 
 
 def connect(connection_name, options):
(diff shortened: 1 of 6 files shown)


src/connectors/duo_collect.py
```diff
@@ -0,0 +1,94 @@
...
+            'subdomain': subdomain,
+            'duo_integration_key': duo_integration_key,
+            'duo_key': duo_key,
+        },
+        dryrun=dryrun,
+    )
+
+
+if __name__ == '__main__':
+    fire(main)
```
sfc-gh-afedorov, line 94: if you want to add connect to CLI, I think best way would be to rename `main` to `cli_ingest` and then add a `cli_connect` that does the same thing, but for the `connect` function. then you can pass them into fire together per these docs — https://github.com/google/python-fire/blob/master/docs/guide.md#version-2-firefiredict

sfc-gh-afedorov:
tested by running in dev, got command —

	CREATE TABLE data.duo_default_admins_connection(raw VARIANT, recorded_at TIMESTAMP_LTZ DEFAULT CURRENT_TIMESTAMP)
	COMMENT='---
	duo_integration_key: asdfasdf
	duo_key: asdfasdf
	module: duo
	name: default
	subdomain: test
	';

and ran these to test rest —

	INSERT INTO data.duo_default_admins_connection (raw) SELECT PARSE_JSON(column1) FROM VALUES ('{}');

	SELECT * FROM data.duo_default_admins_connection;

got result as expected

PR #428, opened 2020-08-13
Greathorn logs

sfc-gh-mpolyakova:


sfc-gh-mpolyakova:
@sfc-gh-afedorov Hey, please take a look when you have a chance! If you have an ETA when you could review, would be helpful. 
Thank you!

src/connectors/greathorn_logs.py
```diff
@@ -0,0 +1,124 @@
...
+LANDING_EVENTS_TABLE_COLUMNS = [('raw', 'VARIANT', 'RECORDED_AT', 'TIMESTAMP_LTZ')]
+
+def connect(connection_name, options):
+    table_name = f'greathorn_{connection_name}'
+    landing_events_table = f'data.{table_name}_events_connection'
+
+    db.create_table(
+        name=landing_events_table,
+        cols=LANDING_EVENTS_TABLE_COLUMNS,
+        comment=yaml_dump(module='duo', **options),
```
sfc-gh-afedorov, line 30: module will refer to the file that does the ingest -- on `greathorn_logs` for this one

src/connectors/greathorn_logs.py
```diff
@@ -0,0 +1,124 @@
...
+
+def ingest(table_name, options, dryrun=False):
+    landing_table = f'data.{table_name}'
+    url = "https://api.greathorn.com/v2/search/events"
+    token = options['api_key']
+
+     starttime = db.fetch_latest(landing_table, 'event_time')
+     if starttime is None:
+         log.error(
+             "Unable to find a timestamp of most recent Okta log, "
```
sfc-gh-afedorov, line 48: Greathorn, not Okta log

src/connectors/greathorn_logs.py
```diff
@@ -0,0 +1,124 @@
...
+
+    while True:
+
+        first_data = json.dumps(
+            {"limit": PAGE_SIZE, "offset": offset, "filters": f_filters}
+        )
+        response = requests.post(url, headers=headers, data=first_data)
+        if response.status_code != 200:
+            log.error(
+             "Response not 200"
```
sfc-gh-afedorov, line 76: use black formatter

src/connectors/greathorn_logs.py
```diff
@@ -0,0 +1,124 @@
...
+        "accept": "application/json",
+        "content-type": "application/json",
+    }
+
+    while True:
+
+        first_data = json.dumps(
+            {"limit": PAGE_SIZE, "offset": offset, "filters": f_filters}
+        )
+        response = requests.post(url, headers=headers, data=first_data)
```
sfc-gh-afedorov, line 73: response can have a `json=<dict>` param which sets headers and lets you skip the json.dumps

src/connectors/greathorn_logs.py
```diff
@@ -0,0 +1,124 @@
...
+            f_filters = [
+                {
+                    "startDate": starttime.strftime("%Y-%m-%d, %H:%M:%S"),
+                    "endDate": endtime.strftime("%Y-%m-%d, %H:%M:%S"),
+                }
+            ]
+            offset = 0
+            yield 0
+
+        events = results['results']
```
sfc-gh-afedorov, line 96: I think there a `db.insert` missing here?

sfc-gh-mpolyakova: Theres definitely a db.insert missing,  sorry about that.

src/connectors/greathorn_logs.py
sfc-gh-mpolyakova: Im not sure. what this means?

sfc-gh-mpolyakova: What should I do for this?

sfc-gh-afedorov: Something like —

    requests.post(url, headers=headers, json={"limit": PAGE_SIZE, "offset": offset, "filters": f_filters})

This will set the headers, as well. Just aesthetic, so no worries if you'd rather not.

src/connectors/greathorn_logs.py
```diff
@@ -0,0 +1,128 @@
...
+        db.insert(
+            f'data.{table_name}',
+            [{'raw': a} for a in results],
```
sfc-gh-afedorov, line 98: did you mean to set `recorded_at` to `default current_timestamp` above, or pass in a single timestamp that is recorded at the start of the script run?

sfc-gh-afedorov (commented):
not sure how you intend to set datetime, then let's grab a 15-30min zoom and test it locally before commit?

src/connectors/greathorn_logs.py
sfc-gh-mpolyakova: I would be cool with either,   probably whichever is convention for connectors.
Which one is more common in connectors?

sfc-gh-afedorov: ideal is to take it from the response — e.g. AWS has a header and Okta includes a "published" time in the entities. second-best is to record time that it's inserted, and the time the script started running is a distant third.

PR #432, opened 2020-08-28
Generic ExtFunc lambda + examples

sfc-gh-afedorov:


infra/extfunc-lambda/lambda_function.py
```diff
@@ -0,0 +1,197 @@
...
+                    data = result
+                    next_url = None
+
+        elif event['path'] == '/smtp':
+            header_params = {
+                k.replace('sf-custom-', '').replace('-', '_'): v.format(*args)
+                for k, v in headers.items()
+                if k.startswith('sf-custom-')
+            }
+            data = smtp(**header_params)
```
sfc-gh-afedorov, line 104: user confused by error
> smtp() got an unexpected keyword argument 'from' Traceback !!

infra/extfunc-lambda/lambda_function.py
```diff
@@ -0,0 +1,197 @@
...
+                    }
+
+                if req_nextpage_path and isinstance(result, list):
+                    data += result
+                    nextpage = pick(req_nextpage_path, result)
+                    next_url = f'https://{req_host}{nextpage}' if nextpage else None
+                elif links_headers and isinstance(result, list):
+                    data += result
+                    next_url = next(
+                        (l for l in links_headers if l['rel'] == 'next'), {}
```
sfc-gh-afedorov, line 92: this loops forever in oktapreview sometimes

infra/extfunc-lambda/lambda_function.py
```diff
@@ -0,0 +1,197 @@
...
+                    ).get('url')
+                else:
+                    data = result
+                    next_url = None
+
+        elif event['path'] == '/smtp':
+            header_params = {
```
sfc-gh-afedorov, line 99: bring up, pass into http method after same transformation

infra/extfunc-lambda/lambda_function.py
```diff
@@ -0,0 +1,197 @@
...
+                k.replace('sf-custom-', '').replace('-', '_'): v.format(*args)
+                for k, v in headers.items()
+                if k.startswith('sf-custom-')
+            }
+            data = smtp(**header_params)
+
+        res_data.append([row, data])
+
+    return {'statusCode': 200, 'body': dumps({'data': res_data})}
```
sfc-gh-afedorov, line 108: if the body variable is too long (>6MB), the code running the lambda will throw an exception. replace it here, instead.

infra/extfunc-lambda/lambda_function.py
```diff
@@ -0,0 +1,204 @@
...
+        req_url = f'https://{req_host}{req_path}'
+
+    req_kwargs = parse_header_dict(headers.get('sf-custom-kwargs'))
+
+    req_headers = {
+        k: v.format(**req_kwargs)
+        for k, v in parse_header_dict(headers.get('sf-custom-headers', '')).items()
+    }
+    if 'sf-custom-basicauth' in headers:
+        basicauth = decrypt_if_encrypted(headers['sf-custom-basicauth'])
```
sfc-gh-afedorov, line 39: the secret should get bundled with an audience and we should check that audience matches the domain before passing the secret off to it

infra/extfunc-lambda/lambda_function.py
```diff
@@ -0,0 +1,204 @@
...
+                    }
+                except JSONDecodeError as e:
+                    result = {
+                        'error': 'JSONDecodeError',
+                        'text': response_body.decode(),
+                    }
+
+                if req_nextpage_path and isinstance(result, list):
+                    data += result
+                    nextpage = pick(req_nextpage_path, result)
```
sfc-gh-afedorov, line 94: when there are too many pages and this runs for >30s, the API Gateway throws an exception. return after 10s here, instead.

infra/extfunc-lambda/lambda_function.py
sfc-gh-afedorov: fixed above in `auth.get('host') != req_host` check

infra/terraform/external_function/snowalert/lambda.tf
```diff
@@ -0,0 +1,60 @@
...
+
+resource "aws_lambda_function" "stdefn" {
+  function_name    = "${var.prefix}_external_function"
+  role             = aws_iam_role.stdefn.arn
+  handler          = "lambda_function.lambda_handler"
+  memory_size      = "512"
+  runtime          = "python3.8"
+  timeout          = "300"
+  publish          = null
+  filename         = "${path.module}/lambda-code.zip"
```
sfc-gh-afedorov, line 20: should this be a variable to `data.archive_file.output_path`?

PR #441, opened 2020-09-23
Task based runners

sfc-gh-afedorov:

PR #442, opened 2020-09-24
feature: Enable Sentry exception tracking

edulop91:
Augmenting the exception tracking with Sentry (instead of only airbrake).

Tested by running
```
AWS_PROFILE=foo SENTRY_DSN=bar python runners/connectors_runner.py "baz"
```


sfc-gh-afedorov (approved):
lgtm. heads up that we'll keep these around for compatibility but WebUI and installer will move to using TASKS + PROCEDURES + JS for alerting

edulop91:
> lgtm. heads up that we'll keep these around for compatibility but WebUI and installer will move to using TASKS + PROCEDURES + JS for alerting

Oh, that sounds great. I'd be happy to write some code in the new framework if that'd be helpful.

sfc-gh-afedorov:
We're still working through the design, but some PoC code that we're aiming to finalize and ship in October is in #432.

The general [External Functional docs](https://docs.snowflake.com/en/sql-reference/external-functions-introduction.html) are helpful to understand the architecture, altho there should be Terraform published with it, as well.

sfc-gh-afedorov:
The EF part is in v1.10.0 branch, but the Tasks-based runners are still WIP (deployed only experimentally) in #441. Let me know if you're curious to migrate to them and we can provide instructions.

PR #447, opened 2020-10-15
Add azic nics

sfc-gh-pkommini:
- Add Network Interfaces to AzIC
- Add `apis` argument to the `runners/connectors_runner.py` script

tested against dev


.gitignore
```diff
@@ -1,20 +1,281 @@
-.DS_Store
```
sfc-gh-afedorov, line 1: let's do one change per commit so as to have separate conversations about separate changes.

.gitignore
```diff
@@ -1,20 +1,281 @@
...
+# Suggestions for gitignore - Python
+# https://github.com/github/gitignore/blob/master/Python.gitignore
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
```
sfc-gh-afedorov, line 10: for this one, I don't think we should copy-paste code without understanding it, so happy to add ignores to something specific you need, but we don't have any C extensions, so the code is vestigial

sfc-gh-afedorov (approved):

PR #449, opened 2020-10-20
Add DC: GCP CloudAssets

sfc-gh-afedorov:
- creates gcp_cloudassets.py with `ingest` method

tested locally against dev/prod

- created gcp_cloudassets.sql.fmt "pack" template

tested using internal code against dev / prod

minor update to db.py tested by automated tests


src/connectors/gcp_cloudassets.py
```diff
@@ -0,0 +1,84 @@
+"""GCP Cloud Asset
+Collect GCP Cloud Assets via API and PIPE
```
sfc-gh-afedorov, line 2: not a pipe

src/connectors/gcp_cloudassets.py
```diff
@@ -0,0 +1,84 @@
...
+LANDING_TABLES_COLUMNS = {
+    'export': [
+        ('created_on', 'TIMESTAMP_LTZ DEFAULT CURRENT_TIMESTAMP'),
+        ('org_id', 'STRING'),
+        ('raw', 'VARIANT'),
+    ]
+}
+
+
+def exec(client, org_id, prefix):
```
sfc-gh-afedorov, line 47: need better name + docstring

sfc-gh-afedorov: `init_asset_export_to_gcs`

src/connectors/gcp_cloudassets.sql.fmt
```diff
@@ -0,0 +1,102 @@
+-- ---
```
sfc-gh-afedorov, line 1: this is confusing without the SQL generating runner

sfc-gh-pkommini (approved):

Issue #453, opened 2020-10-27
s3:get_public_access_block


edulop91:
Hello! I was wondering if there was any interest in adding the [s3:get_public_access_block](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_public_access_block) to the aws collect connector.

If so, I'd be happy to contribute the code :smile:

sfc-gh-afedorov:
Sounds useful, would appreciate, thank you!

sfc-gh-afedorov:
Added it to the v1.10.0 release via PR #459

Could you please double check it's working as intended, @edulop91 ?

edulop91:
Verified, thank you @sfc-gh-afedorov !

PR #454, opened 2020-10-27
[DC] Add aws_collect s3.get_public_access_block

edulop91:
Fixes https://github.com/snowflakedb/SnowAlert/issues/453

Tested by:
- running the migration sql script
- running `python runners/connectors_runner.py "AWS_COLLECT_%%"` locally (under snowalert snowflake role)
- `select * from data.aws_collect_s3_get_public_access_block;` 

One thing to note is that we'll get 404's and a corresponding `ClientError` for buckets with no public access block.

### DIFF
--- a/migrations/v1_9_6-v1_10_0.md
+++ b/migrations/v1_9_6-v1_10_0.md
@@ -12,7 +12,7 @@ CREATE OR REPLACE TABLE azure_collect_network_interfaces (
     location STRING,
     properties VARIANT,
     tags VARIANT,
-    type STRING,
+    type STRING
 );
 GRANT SELECT, INSERT
 ON azure_collect_network_interfaces
@@ -25,12 +25,12 @@ TO ROLE snowalert;
 CREATE OR REPLACE TABLE data.aws_collect_s3_get_public_access_block (
     recorded_at TIMESTAMP_LTZ,
     account_ID STRING,
-    error VARIANT,
     bucket STRING,
+    error VARIANT,
     ignore_public_acls BOOLEAN,
     block_public_policy BOOLEAN,
     block_public_acls BOOLEAN,
-    restrict_public_buckets BOOLEAN,
+    restrict_public_buckets BOOLEAN
 );
 GRANT SELECT, INSERT
 ON data.aws_collect_s3_get_public_access_block
--- a/migrations/v1_9_6-v1_10_0.md
+++ b/migrations/v1_9_6-v1_10_0.md
@@ -18,3 +18,21 @@ GRANT SELECT, INSERT
 ON azure_collect_network_interfaces
 TO ROLE snowalert;
 ```
+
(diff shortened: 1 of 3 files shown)


migrations/v1_9_6-v1_10_0.md
```diff
@@ -18,3 +18,21 @@ GRANT SELECT, INSERT
...
+```sql
+CREATE OR REPLACE TABLE data.aws_collect_s3_get_public_access_block (
+    recorded_at TIMESTAMP_LTZ,
+    account_ID STRING,
+    error VARIANT,
+    bucket STRING,
+    ignore_public_acls BOOLEAN,
+    block_public_policy BOOLEAN,
+    block_public_acls BOOLEAN,
+    restrict_public_buckets BOOLEAN,
```
sfc-gh-afedorov, line 33: trailing comma will break

src/connectors/aws_collect.py
```diff
@@ -459,6 +459,17 @@
         ('target_grants', 'VARIANT'),
         ('target_prefix', 'STRING'),
     ],
+    # https://docs.aws.amazon.com/cli/latest/reference/s3api/get-public-access-block.html
+    's3_get_public_access_block': [
+        ('recorded_at', 'TIMESTAMP_LTZ'),
+        ('account_id', 'STRING'),
+        ('error', 'VARIANT'),
+        ('bucket', 'STRING'),
```
sfc-gh-afedorov, line 467: sorry for inconsistency above, but "bucket" should go before "error", since it's a parameter not a result

src/connectors/aws_collect.py
```diff
@@ -459,6 +459,17 @@
...
+        ('ignore_public_acls', 'BOOLEAN'),
+        ('block_public_policy', 'BOOLEAN'),
```
sfc-gh-afedorov, line 469: maintain same order as in docs, please —

1. BlockPublicAcls
2. IgnorePublicAcls
3. BlockPublicPolicy
4. RestrictPublicBuckets

sfc-gh-afedorov (commented):
please test before merging, but this looks like it should work OK. thanks for the contribution!

src/connectors/aws_collect.py
edulop91: huh, the output example is different from the parameter enumeration. I'll opt for the enumeration like you suggest

sfc-gh-afedorov: Yeah, not sure how that works, but went with treating the enumeration as authoritative since that's what e.g. boto3 does https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3control.html#S3Control.Client.get_public_access_block

sfc-gh-afedorov (approved):
lgtm

sfc-gh-pkommini:
Ran local tests. Works fine. lgtm.

PR #457, opened 2020-11-02
[DC]: Add Accountwide S3 Bucket Public Access Block

edulop91:
Related to https://github.com/snowflakedb/SnowAlert/pull/454 and https://github.com/snowflakedb/SnowAlert/issues/453

Pulls the account level S3 Bucket Public Access.
There is something a bit wonky where the api requires me to pass in the accountid (of the account i'm calling this from). Seemed like the cleanest solution was to also chain an sts.get_caller_identity to rely on the parameter propagation. Please let me know if there is a better way. Also kind of confusing that the api call is `s3control`, not intuitive that this is an account wide resource.

Tested by:
1. `python runners/connectors_runner.py "AWS_COLLECT_%%"`
1. `select * from "SNOWALERT"."DATA"."AWS_COLLECT_S3CONTROL_GET_PUBLIC_ACCESS_BLOCK";`
1. `select * from "SNOWALERT"."DATA"."AWS_COLLECT_STS_GET_CALLER_IDENTITY";`

### DIFF
--- a/src/connectors/aws_collect.py
+++ b/src/connectors/aws_collect.py
@@ -470,6 +470,16 @@ SUPPLEMENTARY_TABLES = {
         ('block_public_policy', 'BOOLEAN'),
         ('restrict_public_buckets', 'BOOLEAN'),
     ],
+    # https://docs.aws.amazon.com/cli/latest/reference/s3control/get-public-access-block.html
+    's3control_get_public_access_block': [
+        ('recorded_at', 'TIMESTAMP_LTZ'),
+        ('account_id', 'STRING'),
+        ('error', 'VARIANT'),
+        ('block_public_acls', 'BOOLEAN'),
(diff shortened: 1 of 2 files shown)


src/connectors/aws_collect.py
```diff
@@ -562,6 +572,14 @@
         ('created_at', 'TIMESTAMP_NTZ'),
         ('updated_at', 'TIMESTAMP_NTZ'),
     ],
+    # https://docs.aws.amazon.com/cli/latest/reference/sts/get-caller-identity.html
+    'sts_get_caller_identity': [
+        ('recorded_at', 'TIMESTAMP_LTZ'),
+        ('account_id', 'STRING'),
+        ('account', 'STRING'),
```
edulop91, line 579: maybe silly to duplicate these, but kept for convention

sfc-gh-afedorov:
merged in #459

PR #460, opened 2020-11-06
Add format button on RawEditor

sfc-gh-pkommini:
Created format button


sfc-gh-afedorov:
- add description re: how it was tested
- what's up with the reformatting? I think your editor might be reverting styling that the build step makes

sfc-gh-afedorov (approved):
lgtm

PR #461, opened 2020-11-06
update aws_query_pack to match GRANT statement

asproul:
* Match GRANT SELECT statement to created rule above it to match convention.
* ~remove cloudtrail.deployment as it is not referenced in [aws_cloudtrail.py](https://github.com/snowflakedb/SnowAlert/blob/v1.9.6/src/connectors/aws_cloudtrail.py) when creating new tables.~

### DIFF
--- a/Dockerfile.snowalert
+++ b/Dockerfile.snowalert
@@ -3,6 +3,8 @@ FROM python:3.7-slim-stretch
 WORKDIR /var/task
 
 RUN apt-get update
+RUN apt-get install -y r-base
+RUN R -e "install.packages(c('dplyr', 'purrr', 'tidyr','MASS', 'tidyverse', 'broom','testthat'), repos = 'https://cloud.r-project.org')"
 
 RUN pip install --upgrade pip virtualenv pyflakes
 
--- a/Dockerfile.snowalert
+++ b/Dockerfile.snowalert
@@ -3,8 +3,6 @@ FROM python:3.7-slim-stretch
 WORKDIR /var/task
 
 RUN apt-get update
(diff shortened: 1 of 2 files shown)


sfc-gh-afedorov:
Thanks for the PR. Could you please expound on what you mean by —

> remove cloudtrail.deployment as it is not referenced in aws_cloudtrail.py when creating new tables.

It can be referenced by some other piece of infra that is built on top of it.

asproul:
> Thanks for the PR. Could you please expound on what you mean by —
> 
> > remove cloudtrail.deployment as it is not referenced in aws_cloudtrail.py when creating new tables.
> 
> It can be referenced by some other piece of infra that is built on top of it.

I could be misunderstanding the "cloudtrail.deployment" construct but, I did not see a "deployment" column for the landing table [created by aws_cloudtrail.py](https://github.com/snowflakedb/SnowAlert/blob/master/src/connectors/aws_cloudtrail.py#L194-L202). When I tried to create the view and run the query as-is in the snowflake worksheet from a newly created landing table using SnowAlert's connector for cloudtrail, the query produced an error stating that "cloudtrail.deployment" did not exist or something to that effect. When searching through this repository, I did not find any references to deployment except the hits in the "aws_cloudtrail.py" file. I thought that "cloudtrail.deployment" might have been a column in an older version of the landing table so I had removed it in my case.

asproul:
@sfc-gh-afedorov  - added `cloudtrail.deployment` back to the file. How's this change looking now?

sfc-gh-afedorov (approved):
lgtm, ty for fixes

PR #464, opened 2020-11-07
Support oauth for snowflake deployments in non-default regions

asproul:
Construct the snowflake hostname i.e. `{account}.snowflakecomputing.com` or `{account}.{region}.snowflakecomputing.com` following the [snowflake documentation](https://docs.snowflake.com/en/user-guide/oauth-custom.html#authorization-endpoint) for multiple regions. Inspired by https://github.com/snowflakedb/snowflake-connector-python/blob/e2d835d6e574e7e1856e60e703d9b5f8826e369a/src/snowflake/connector/connection.py#L686-L688 

Supplying the region to connect_db allows one to connect to the region specific endpoint for snowflake. Tested this change by building and running the web-ui docker image, I was able to login using oauth to the SnowAlert web ui to my snowflake account located in us-east-1.

Before these changes, I was not able to login using oauth as the web ui was hitting the default region endpoint (mydemo.snowflakecomputing.com) rather than (mydemo.us-east-1.snowflakecomputing.com).

### DIFF
--- a/src/webui/backend/webui/api/oauth.py
+++ b/src/webui/backend/webui/api/oauth.py
@@ -7,6 +7,10 @@ import requests
 from flask import Blueprint, request, jsonify
 import logbook
 
+from snowflake.connector.util_text import (
+    construct_hostname as construct_snowflake_hostname,
+)
+
 logger = logbook.Logger(__name__)
 
 oauth_api = Blueprint('oauth', __name__)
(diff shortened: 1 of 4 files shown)


src/webui/backend/webui/api/oauth.py
```diff
@@ -35,8 +38,10 @@ def oauth_redirect():
     else:
         scope_role = ''
 
+    hostname = construct_snowflake_hostname(REGION, account)
```
sfc-gh-afedorov, line 41: did you test this with WebUI? I think this is currently supported by a user entering the region to complete the URL in the login field, here —

> image

changing it like this will break that support, tho, and limit the UI to only work with a single region's accounts. the idea right now is that a single WebUI deployment should allow folks to log into two accounts in two different regions.

sfc-gh-afedorov (changes requested):
have you tried entering the non-default region after a ".", e.g. "ab1235.us-east-1" on the login page? that is the intended usage for this feature

src/webui/backend/webui/api/oauth.py
asproul: Ah. I had not tried that. Will close this PR. Thank you.

PR #473, opened 2020-12-17
increase limit when populating users in an okta group

sfc-gh-dwyleczukstern:
Default was 1000

https://developer.okta.com/docs/reference/api/groups/#list-group-members

Based syntax on that - I think I did it right but haven't gotten all the tokens setup to see

### DIFF
--- a/src/connectors/okta.py
+++ b/src/connectors/okta.py
@@ -81,6 +81,13 @@ def connect(connection_name, options):
         'newMessage': "Okta ingestion table, user table, group table created!",
     }
 
+def get_next_page(response):
+    url = ''
+    links = requests.utils.parse_header_links(response.headers['Link'])
+    for link in links:
+        if link['rel'] == 'next':
+            url = link['url']
+    return url
 
 def ingest_users(url, headers, landing_table, now):
     while 1:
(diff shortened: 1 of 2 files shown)


src/connectors/okta.py
```diff
@@ -146,19 +146,33 @@ def ingest(table_name, options):
...
+            while has_next_page:    
+                try:
+                    page = requests.get(
+                        url=url, headers=headers
+                    )
+                    users += page.json()
+                    if 'rel="next"' in page.headers['link']:
+                        for item in str.split(page.headers['link'], ','):
+                            if 'rel="next"' in item:
+                                url = re.findall(r'(?<=\<)(.*?)(?=\>)', item)[0]
```
sfc-gh-afedorov, line 161: could you use same code block as above, and maybe make it into a helper function?

        url = ''
        links = requests.utils.parse_header_links(response.headers['Link'])
        for link in links:
            if link['rel'] == 'next':
                url = link['url']

sfc-gh-dwyleczukstern: Change made as requested

sfc-gh-afedorov (approved):

sfc-gh-pkommini:
@sfc-gh-dwyleczukstern Sorry I missed this since December. Do we still want to merge this?

sfc-gh-pkommini (approved):
lgtm.

PR #475, opened 2020-12-18
Add h1 data connector

sfc-gh-pkommini:
**Changes**
- Add H1 data connectors

**Tests**
- Tested against dev
- Tested against staging
- Tested `connect` using dev webui
image


src/connectors/h1_collect.py
```diff
@@ -0,0 +1,316 @@
...
+    ('name', 'string'),
+    ('description', 'string'),
+    ('external_id', 'string'),
+    ('asset_identifier', 'string'),
+    ('awarded_amount', 'int'),
+    ('awarded_bonus_amount', 'int'),
+]
+
+
+# Perform the API call
```
sfc-gh-afedorov, line 85: not sure if this is necessary with more specific name, but if you'd like to include it it's better in [a docstring](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)

src/connectors/h1_collect.py
```diff
@@ -0,0 +1,316 @@
...
+    headers: dict = {"Accept": "application/json"}
+
+    try:
+        log.debug(f"Preparing GET: url={url} with params={params}")
+        resp = requests.get(
+            url, auth=(api_identifier, token), params=params, headers=headers
+        )
+        resp.raise_for_status()
+    except HTTPError as http_err:
+        log.error(f"Error GET: url={url}")
```
sfc-gh-afedorov, line 96: does the exception not contain the URL? even if not, seems we log it via `log.debug` above so this is vestigial. please remove.

src/connectors/h1_collect.py
```diff
@@ -0,0 +1,316 @@
...
+    if ingest_type == 'transaction' and account_id:
+        try:
+            transactions = load_data(
+                f'https://api.hackerone.com/v1/programs/{account_id}/billing/transactions',
+                api_token,
+                api_identifier,
+                params=None,
+            )
+        except requests.exceptions.HTTPError as e:
+            log.error(e)
```
sfc-gh-afedorov, line 160: this will log an error to CloudWatch, but not record it in the metadata tables. the runners that calls into this expects exceptions to be raised if there is an error and handle them [here](https://github.com/snowflakedb/SnowAlert/blob/master/src/runners/connectors_runner.py#L107).

in general, try / except should be used in code that decides how to handle errors, and so we want to do it in a general way by the runner framework rather than in each handler.

src/connectors/h1_collect.py
```diff
@@ -0,0 +1,316 @@
...
+            db.insert(
+                landing_table,
+                dryrun=dryrun,
+                values=[
+                    {
+                        'timestamp': timestamp,
+                        'id': report.get('id', ''),
+                        'type': report.get('type', ''),
+                        'title': report['attributes'].get('title', '')
+                        if 'attributes' in report
```
sfc-gh-afedorov, line 209: this is a bit rough to read -- could we use the `get_path` util I added above?

src/connectors/h1_collect.py
```diff
@@ -0,0 +1,316 @@
...
+        except requests.exceptions.HTTPError as e:
+            log.error(e)
+        else:
+            db.insert(
+                landing_table,
+                dryrun=dryrun,
+                values=[
+                    {
+                        'timestamp': timestamp,
+                        'id': report.get('id', ''),
```
sfc-gh-afedorov, line 206: `id` col is defined as number above so I don't think you can set it to `''` without raising an error

src/connectors/h1_collect.py
```diff
@@ -0,0 +1,316 @@
...
+                        'type': report.get('type', ''),
+                        'title': report['attributes'].get('title', '')
+                        if 'attributes' in report
+                        else '',
+                        'state': report['attributes'].get('state', '')
+                        if 'attributes' in report
+                        else '',
+                        'created_at': report['attributes'].get('created_at', '')
```
sfc-gh-afedorov, line 214: this is a `timestamp_ltz` above so I think setting it to `''` will raise an error

src/connectors/h1_collect.py
```diff
@@ -0,0 +1,316 @@
...
+    for key in path.split("."):
+        if key not in data:
+            return default
+        else:
+            data = data[key]
+    return data
+
+
+def connect(connection_name, options):
+    landing_transactions_table = f'data.h1_collect_transactions_connection'
```
sfc-gh-afedorov, line 113: use the `connection_name` here so that users can make more than one h1 connection should they choose to —

    landing_transactions_table = f'data.h1_collect_{connection_name}_transactions_connection'

src/connectors/h1_collect.py
```diff
@@ -0,0 +1,316 @@
...
+
+    comment = yaml_dump(module='h1_collect', **options)
+
+    db.create_table(
+        name=landing_transactions_table,
+        cols=LANDING_TABLE_COLUMNS_TRANSACTIONS,
+        comment=comment,
+    )
+    db.execute(
+        f'GRANT INSERT, SELECT ON {landing_transactions_table} TO ROLE {SA_ROLE}'
```
sfc-gh-afedorov, line 124: iirc, this can be passed as `rw_role` to `create_table`

src/connectors/h1_collect.py
```diff
@@ -0,0 +1,316 @@
...
+        resp.raise_for_status()
+    except HTTPError as http_err:
+        log.error(f"Error GET: url={url}")
+        log.error(f"HTTP error occurred: {http_err}")
+        raise
+    log.debug(resp.status_code)
+    return resp.json()['data']
+
+
+def get_path(data, path, default=None):
```
sfc-gh-pkommini, line 103: @sfc-gh-afedorov Do you want to move this to utils? So that it can be used globally. Seems handy.

sfc-gh-afedorov (changes requested):
nice!

the only blocking errors are the exception handling and the uncertainty about column types. in general, having `None` instead of `''` for missing values is preferred.

src/connectors/h1_collect.py
sfc-gh-afedorov: no strong feelings as to when, but usually wait until it's useful in a second place to "promote" it to the utils. in this case could see it being useful ahead of time, so either way works.

src/connectors/h1_collect.py
sfc-gh-pkommini: ACK.

src/connectors/h1_collect.py
sfc-gh-ttruong: We copied the template from merkai_devices.py (lines80-86), would you like it rectified there too?

src/connectors/h1_collect.py
sfc-gh-ttruong: To clarify, we should remove the try/except (lines 151-159) and let the runner handle the errors? Or should we do a raise at line 160 to pass the error to the runners as well? ie where should the error be logged?

src/connectors/h1_collect.py
sfc-gh-ttruong: Is None or 0 an acceptable default to pass here then?

src/connectors/h1_collect.py
sfc-gh-ttruong: Do we remove the 2 tables (`transactions` and `reports`) that we've created then and replace it with one generic table with `{connection_name}`? Also, when we click `connect`, will it automatically do it for all the tables (`transactions` and `reports`) associated that we need?

src/connectors/h1_collect.py
sfc-gh-afedorov: yes, altho iirc we don't have any Markai devices to test changes against so not urgent

src/connectors/h1_collect.py
sfc-gh-afedorov: up to you, but I would just remove. the runners will log the error.

src/connectors/h1_collect.py
sfc-gh-afedorov: I would say `None` as `0` might be an actual `id` (unlikely, but the very first email for a customer, maybe?)

src/connectors/h1_collect.py
sfc-gh-ttruong: Do we remove lines 122-124 and replace it with `rw_role=SA_ROLE` in the `create_table` call?

src/connectors/h1_collect.py
sfc-gh-afedorov: leave the two tables that are there now, but create three new tables using this code. I think it would be OK to do —

    data.h1_collect__events_connection
    data.h1_collect__transactions_connection
    data.h1_collect__report_connection

or replace the `__` with `_` like some of the others do, i.e. with —

    landing_transactions_table = f'data.h1_collect_{connection_name}_transactions_connection'.replace('__', '_')

src/connectors/h1_collect.py
sfc-gh-afedorov: Yes, like [here](https://github.com/snowflakedb/SnowAlert/blob/40f38a3f7fbd9c3cde7e7e51dd72ae881a70cd0c/src/connectors/airwatch_devices.py#L148-L153).

src/connectors/h1_collect.py
```diff
@@ -0,0 +1,251 @@
...
+        log.debug(f"Preparing GET: url={url} with params={params}")
+        resp = requests.get(
+            url, auth=(api_identifier, token), params=params, headers=headers
+        )
+        resp.raise_for_status()
+    except HTTPError as http_err:
+        log.error(f"HTTP error occurred: {http_err}")
+        raise
+    log.debug(resp.status_code)
+    return resp
```
sfc-gh-afedorov, line 100: Merged in changes on v1.10.0 so that we can run the automated tests which [reported](https://app.circleci.com/pipelines/github/snowflakedb/SnowAlert/426/workflows/cb8b86ff-48be-47f9-b3cc-a8399b885c7f/jobs/4550):

    connectors/h1_collect.py:100: error: Incompatible return value type (got "Response", expected "Dict[Any, Any]")

sfc-gh-pkommini: Fixed the issue. It was the return type hint. Changed that to `requests.Response`

src/connectors/h1_collect.py
```diff
@@ -0,0 +1,256 @@
...
+    ('awarded_bonus_amount', 'int'),
+]
+
+
+def load_data(
+    url: str,
+    token: str,
+    api_identifier: str,
+    params: dict = {}
+) -> requests.Response:
```
sfc-gh-pkommini, line 91: @sfc-gh-afedorov: This change should fix the error shown by mypy.

sfc-gh-afedorov (approved):

PR #476, opened 2020-12-18
[DC] Add S3 Get Bucket Tagging

edulop91:
ran it locally and made sure data populated in snowflake appropriately

### DIFF
--- a/src/connectors/aws_collect.py
+++ b/src/connectors/aws_collect.py
@@ -459,6 +459,14 @@ SUPPLEMENTARY_TABLES = {
         ('target_grants', 'VARIANT'),
         ('target_prefix', 'STRING'),
     ],
+    # https://docs.aws.amazon.com/cli/latest/reference/s3api/get-bucket-tagging.html#output
+    's3_get_bucket_tagging': [
+        ('recorded_at', 'TIMESTAMP_LTZ'),
+        ('account_id', 'STRING'),
+        ('error', 'VARIANT'),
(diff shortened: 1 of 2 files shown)


src/connectors/aws_collect.py
```diff
@@ -459,6 +459,14 @@
         ('target_grants', 'VARIANT'),
         ('target_prefix', 'STRING'),
     ],
+    # https://docs.aws.amazon.com/cli/latest/reference/s3api/get-bucket-tagging.html#output
+    's3_get_bucket_tagging': [
+        ('recorded_at', 'TIMESTAMP_LTZ'),
+        ('account_id', 'STRING'),
+        ('error', 'VARIANT'),
+        ('bucket', 'STRING'),
```
edulop91, line 467: I think I keep messing up the order of these

edulop91: although error then bucket seems to be the more common one

sfc-gh-afedorov: all good, it's an unfortunate limitation. the params go before error, and then either "error" or everything after it is populated with response. if all of them are NULL then there was no error and the response was empty.

sfc-gh-afedorov: bucket should be after error if the bucket is part of the response and before the error if it's part of the query

edulop91: thank you

sfc-gh-afedorov (approved):

PR #483, opened 2021-01-08
Add pagination h1 collect

sfc-gh-ttruong:
**_Changes_**:
- Add paginated inserts for reports and transactions
- Change prompt to add context to the URL
- Modified PAGE_SIZE to 100

**_Tests_**:
- Tested in dev and found that records were inserted successfully
- tested with a page size of 1 and it correctly inserted 1 record per page for a total of 2 records


sfc-gh-afedorov (approved):
lgtm

PR #485, opened 2021-01-20
[DC] Add EFS Collection to AWSIC

sfc-gh-dwyleczukstern:


sfc-gh-afedorov (approved):
tested & updated a few minor fixes. the latter did not work without at least one param, so I picked the id off of the former.

PR #486, opened 2021-01-23
Fix amount and transactions

sfc-gh-ttruong:
_**Changes**_
- Add for loop for transactions table
- Create function for awarded_amount and awarded_bonus_amount due to issue with for loop in variable 

**_Test_**
- Prasanth ran data connector test and we were able to produce desired results


src/connectors/h1_collect.py
```diff
@@ -202,73 +209,50 @@ def paginated_insert_reports(landing_table, options, dryrun):
...
-        api_token,
-        api_identifier,
-        params={
-            'filter[program][]': program_name,
-            'page[size]': PAGE_SIZE,
-            'page[number]': page_number
-        },
-    )
-    reports = response.json()['data']
-    recorded_at = parsedate_to_datetime(response.headers.get('Date'))
```
sfc-gh-afedorov, line 217: let's keep this, and move it to the `load_data` function so that it doesn't accidentally get dropped again

src/connectors/h1_collect.py
```diff
@@ -161,14 +170,12 @@ def insert_reports(landing_table, reports, recorded_at, dryrun):
                 'description': get_path(report, 'relationships.weakness.data.attributes.description'),
                 'external_id': get_path(report, 'relationships.weakness.data.attributes.external_id'),
                 'asset_identifier': get_path(report, 'relationships.structured_scope.data.attributes.asset_identifier'),
-                'awarded_amount': sum(
```
sfc-gh-afedorov, line 164: this change looks a bit like a red herring — is this what fixed the bug, or the change below?

src/connectors/h1_collect.py
```diff
@@ -92,17 +109,8 @@ def load_data(
...
-        log.debug(f"Preparing GET: url={url} with params={params}")
-        resp = requests.get(
-            url, auth=(api_identifier, token), params=params, headers=headers
-        )
-        resp.raise_for_status()
-    except HTTPError as http_err:
-        log.error(f"HTTP error occurred: {http_err}")
-        raise
     log.debug(resp.status_code)
-    return resp
```
sfc-gh-afedorov, line 105: you'll need to either skip the type declaration on the function or update it to match the return type. you can either set up mypy to run locally, or see ci/cd results [here](https://app.circleci.com/pipelines/github/snowflakedb/SnowAlert?branch=fix_amount_and_transactions)

src/connectors/h1_collect.py
```diff
@@ -202,73 +209,50 @@ def paginated_insert_reports(landing_table, options, dryrun):
...
-        params={
-            'page[size]': PAGE_SIZE,
-            'page[number]': page_number
-        },
-    )
-    transactions = response.json()['data']
-    recorded_at = parsedate_to_datetime(response.headers.get('Date'))
-    insert_transactions(landing_table, transactions, recorded_at, dryrun)
-
-    while 'next' in response.json()['links']:
```
sfc-gh-afedorov, line 258: am I reading right that this loads a single page from each month? does this API not have pagination like the other one?

sfc-gh-afedorov (changes requested):

src/connectors/h1_collect.py
```diff
@@ -202,73 +209,50 @@ def paginated_insert_reports(landing_table, options, dryrun):
...
+                api_token,
+                api_identifier,
+                params={
+                    'page[size]': PAGE_SIZE,
+                    'month': month,
+                    'year': year
+                },
+            )
+            transactions = response.json()['data']
+            insert_transactions(landing_table, transactions, recorded_at, dryrun)
```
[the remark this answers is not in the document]
sfc-gh-ttruong: I'm not sure how we're adding a new timestamp onto the mix. `recorded_at = datetime.utcnow()` in the entire script.

src/connectors/h1_collect.py
sfc-gh-ttruong: This fixed the issue of awarded_amount and awarded_bonus_amount showing up for only the last page.

src/connectors/h1_collect.py
sfc-gh-ttruong: correct - per https://api.hackerone.com/core-resources/#programs-get-payment-transactions "If you want to get transactions for an entire year, you will need to request each month individually."

src/connectors/h1_collect.py
sfc-gh-afedorov: why do you think so? do you see a logic difference between the two versions I am missing?

src/connectors/h1_collect.py
sfc-gh-afedorov: got it -- that might be worth linking for the explanation. it sounds like the resources are paginated, also, so it would likely make sense to move the pagination logic into the `load_data` function

src/connectors/h1_collect.py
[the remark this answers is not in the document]
sfc-gh-ttruong: Also, I'm already accounting for the next page in def paginated_insert_reports function.

sfc-gh-afedorov (approved):

PR #489, opened 2021-02-09
feat[spotlight]: adding connector for crowdstrike spotlight API

jacobrheath:


src/connectors/crowdstrike_devices.py
```diff
@@ -256,8 +308,258 @@ def ingest(table_name, options):
...
         log.info(f'Inserted {len(devices)} rows.')
         yield len(devices)
+
+
+rem_id_cache = {}
+
+
+async def does_rem_exist(rem_id: str) -> str:
+    """
+    does_rem_exist returns "" if the remediation is has already been cached locally
```
edulop91, line 323:
```suggestion
    does_rem_exist returns "" if the remediation id has already been cached locally
```

src/connectors/crowdstrike_devices.py
```diff
@@ -161,21 +182,52 @@ def create_url_params_get_devices(url: str, resources: list) -> str:
...
+
+    db.create_table(
+        name=landing_spotlight_rems_table,
+        cols=LANDING_SPOTLIGHT_REMS_TABLE_COLUMNS,
+        comment=comment,
+    )
+
+    db.execute(
+        f'GRANT INSERT, SELECT ON {landing_spotlight_rems_table} TO ROLE {SA_ROLE}'
+    )
```
edulop91, line 221:
```suggestion
    db.create_table(
        name=landing_devices_table, cols=LANDING_DEVICES_TABLE_COLUMNS, comment=comment
    )
    db.execute(f'GRANT INSERT, SELECT ON {landing_devices_table} TO ROLE {SA_ROLE}')

    db.create_table(
        name=landing_spotlight_vulns_table,
        cols=LANDING_SPOTLIGHT_VULNS_TABLE_COLUMNS,
        comment=comment,
    )
    db.execute(
        f'GRANT INSERT, SELECT ON {landing_spotlight_vulns_table} TO ROLE {SA_ROLE}'
    )

    db.create_table(
        name=landing_spotlight_rems_table,
        cols=LANDING_SPOTLIGHT_REMS_TABLE_COLUMNS,
        comment=comment,
    )
    db.execute(
        f'GRANT INSERT, SELECT ON {landing_spotlight_rems_table} TO ROLE {SA_ROLE}'
    )
```

src/connectors/crowdstrike_devices.py
```diff
@@ -256,8 +308,258 @@ def ingest(table_name, options):
...
+    or is recorded in the database. Otherwise, it returns the remediation ID.
+    """
+    # Check local cache if you have seen this ID in this run
+    if rem_id in rem_id_cache:
+        log.info(f"[CACHE]: Local cache hit! Already recorded {rem_id} in this round")
+        return ""
+
+    # Check if we have it stored in the table already
+    row = db.fetch_latest(
+        "data.CROWDSTRIKE_SPOTLIGHT_REMS_DEFAULT_CONNECTION",
```
edulop91, line 333: This probably needs to be parameterized like [these](https://github.com/snowflakedb/SnowAlert/pull/489/files#diff-0f3bcae7b198de7ad264a1810da509be9215dbd8427ab4501feb5352144219fdR185)

src/connectors/crowdstrike_devices.py
```diff
@@ -256,8 +308,258 @@ def ingest(table_name, options):
...
+        reduce(
+            (lambda x, y: x + y if (y != None) else x),
+            [det.get("remediation", {}).get("ids", []) for det in vuln_dets],
+            [],
+        )
+    )
+    # find the IDs that we have not already recorded
+    rem_ids = await asyncio.gather(
+        *[asyncio.create_task(does_rem_exist(r)) for r in rem_ids]
+    )
```
edulop91, line 360: This looks like an "n+1 query", given that you have the whole set of rem_ids would it be possible to issue a query to look for the whole set in a batch?

src/connectors/crowdstrike_devices.py
```diff
@@ -256,8 +308,258 @@ def ingest(table_name, options):
...
+
+async def insert_rem_dets(dets: list) -> Coroutine[None, None, int]:
+    """
+    insert_rem_dets inserts a remediation into the database.
+    """
+    if len(dets) == 0:
+        return 0
+
+    numRows = db.insert(
+        f'data.CROWDSTRIKE_SPOTLIGHT_REMS_DEFAULT_CONNECTION',
```
edulop91, line 384: same here about parameterizing the table name

src/connectors/crowdstrike_devices.py
```diff
@@ -256,8 +308,258 @@ def ingest(table_name, options):
...
+def paginate_api(
+    params: Dict[str, str],
+    url: str,
+    max_page_entries: int = 400,
+):
+    """
+    paginate_api returns a generator that yields page entries from the Crowdstrike
+    Spotlight API. By default, page sizes are only 400 entries long because that maps
+    to the maximim number of vuln IDs that can be specified in the vuln details API, but
+    this is configurable.
```
edulop91, line 482:
```suggestion
    paginate_api returns a generator that yields page entries from the specified Crowdstrike
    API. By default, page sizes are only 400 entries long because that maps
    to the maximim number of vuln IDs that can be specified in the vuln details API, but
    this is configurable.
```

src/connectors/crowdstrike_devices.py
edulop91: for spacing consistency

edulop91 (commented):
some small feedback!

src/connectors/crowdstrike_devices.py
jacobrheath: Yup, good call. Fixed this

src/connectors/crowdstrike_devices.py
jacobrheath: So are we not worrying about SQL injection here? Looks like the DB library has a prepared statements method but I don't see many people using it.

jacobrheath: Also, do you know how the connection name stuff works? I only saw the string "DEFAULT", but are there other strings?

sfc-gh-afedorov:
have you tested / are you running this? happy to have it, but not sure if we need it ourselves. any extra explanation of what this data is useful for or instructions for setting it up you can put into the /docs/connectors repo would be appreciated

jacobrheath:
Hey there, at CZI we started using this API to pull in metrics and thought it might be nice to have it as a part of the existing Cloudstrike connector. It's been a while since i have reviewed this code, so I would probably want to go back and test it. I don't think we have started to use the connector yet.

jacobrheath:
If it helps, I can convert to draft until I have more details

jacobrheath:
I think we ended up not really needing this as much, so I am going to close this for now.

jacobrheath:
Turns out we ARE using this data pretty regularly. I am going to clean up this PR and write some tests.

PR #490, opened 2021-02-23
Google Collect Connector

sfc-gh-afedorov:
ran against gcloud sandbox + snowhouse, verified results are landing in `data.google_collect_connection`

### DIFF
--- a/src/connectors/google_collect.py
+++ b/src/connectors/google_collect.py
@@ -1,0 +1,105 @@
+"""Google Collect
+Collect Google API responses using a Service Account
+"""
+
+from collections import namedtuple
+
+from googleapiclient.discovery import build
+from google.oauth2 import service_account
+
+from runners.helpers import db
+from runners.helpers.dbconfig import ROLE as SA_ROLE
+
+from .utils import yaml_dump
+
+CONNECTION_OPTIONS = [
+    {
+        'name': 'service_user_creds',
+        'type': 'json',
+        'title': "Service User Credentials",
+        'prompt': "JSON Credentials from the Service User",
+        'placeholder': (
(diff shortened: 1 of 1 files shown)


src/connectors/google_collect.py
```diff
@@ -0,0 +1,99 @@
...
+]
+
+API_SPECS = {
+    'users': {
+        'service_name': 'admin',
+        'service_version': 'directory_v1',
+        'scopes': ['https://www.googleapis.com/auth/admin.directory.user.readonly'],
+        'resource_name': 'users',
+        'method': 'list',
+        'params': lambda subject: {'domain': subject.split('@')[1]},
```
sfc-gh-afedorov, line 42: this mapping seems like it doesn't belong here

src/connectors/google_collect.py
```diff
@@ -0,0 +1,99 @@
...
+                'method': spec['method'],
+                'params': spec['params'](subject=subject),
+            }
+            response = make_call(service_user_creds, **call_params)
+            db.insert(
+                f'data.{table_name}',
+                {'api_name': api_name, 'response': response, 'request': call_params},
+                dryrun=dryrun,
+            )
+            yield 1
```
sfc-gh-pkommini, line 99: @sfc-gh-afedorov Is `ingest()` a generator function? Is there a reason to use `yield` instead of return in this context?

sfc-gh-afedorov: the runner inspects `ingest` and if it's a generator it iterates through it and sums the yield's — https://github.com/snowflakedb/SnowAlert/blob/master/src/runners/connectors_runner.py#L23

sfc-gh-pkommini (approved):
lgtm

PR #491, opened 2021-02-24
AWS_Cloudtrail_pack

sfc-gh-srathod:
Please review the aws_cloudtrail sql pack

### DIFF
--- a/packs/AWS_CloudTrail.sql
+++ b/packs/AWS_CloudTrail.sql
@@ -1,0 +1,201 @@
+-- varmap:
+--   base_name: 'AWS_CLOUDTRAIL_{connection_name}_EVENTS'
+--   stage: 'data.{base_name}_STAGE'
+--   staging_table: 'data.{base_name}_STAGING'
+--   landing_table: 'data.{base_name}_CONNECTION'
+--   pipe = 'data.{base_name}_PIPE'
+--   task: 'data.{base_name}_TASK'
+--   stream: 'data.{base_name}_STREAM'
+--   storage_integration: 'data.{base_name}_STORAGE_INTEGRATION'
+
+
+CREATE SEQUENCE IF NOT EXISTS LANDING_TABLE_SEQ START 1 INCREMENT 1;
+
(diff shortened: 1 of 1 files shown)


packs/AWS_CloudTrail.sql
```diff
@@ -0,0 +1,201 @@
+-- varmap:
```
anfedorov, line 1: please update header to include params definition

packs/AWS_CloudTrail.sql
```diff
@@ -0,0 +1,201 @@
...
+--   stage: 'data.{base_name}_STAGE'
+--   staging_table: 'data.{base_name}_STAGING'
+--   landing_table: 'data.{base_name}_CONNECTION'
+--   pipe = 'data.{base_name}_PIPE'
+--   task: 'data.{base_name}_TASK'
+--   stream: 'data.{base_name}_STREAM'
+--   storage_integration: 'data.{base_name}_STORAGE_INTEGRATION'
+
+
+CREATE SEQUENCE IF NOT EXISTS LANDING_TABLE_SEQ START 1 INCREMENT 1;
```
anfedorov, line 12: let's keep the casing consistent -- we want our names to be lower case, so `LANDING_TABLE_SEQ` should be, as well, to separate it from the language constructs

packs/AWS_CloudTrail.sql
```diff
@@ -0,0 +1,201 @@
...
+    ON {staging_table}
+;
+
+
+
+--Pipe
+CREATE OR REPLACE PIPE {pipe}
+    AUTO_INGEST = TRUE
+    AS
+    COPY INTO {staging_table(v)}
```
anfedorov, line 101: this is not a valid param

anfedorov (changes requested):
good start, but there's a few pieces left before we can test this

sfc-gh-afedorov:
now in https://github.com/snowflakedb/snowsql-pack/tree/main/aws_cloudtrail

PR #492, opened 2021-02-25
Add groups and members query

sfc-gh-pkommini:
- Add groups
- Add members as child resource to groups


sfc-gh-afedorov:
and latest code ran ok?

sfc-gh-afedorov (approved):
lgtm as long as it runs sans error + add the missing an "s" :-)

PR #494, opened 2021-03-03
collect more GCP data

sfc-gh-dwyleczukstern:


src/connectors/gcp_cloudassets.py
```diff
@@ -81,4 +81,8 @@ def ingest(table_name, options):
         org_id, location = org_location.split(':')
         dt = datetime.utcnow().strftime('%Y/%m/%d/%H:%M:%S')
         prefix = 'gs://' + location + '/cloudassets/' + dt
```
sfc-gh-afedorov, line 83: the prefix needs to vary for each of these calls, too

sfc-gh-dwyleczukstern: Why does the prefix need to vary? This is the location that data gets output too, correct? https://cloud.google.com/asset-inventory/docs/reference/rpc/google.cloud.asset.v1#google.cloud.asset.v1.OutputConfig Would adding another folder break ingestion?

sfc-gh-afedorov: PK tested it and got an error —

> Invalid output GCS file path prefix in request: permission denied or file with same name already exists

I think the ingestion should be Ok if we vary the date or put the name before the date. that code uses two regex to find the type and date —

~~~
  log_type STRING AS (
    REGEXP_REPLACE( -- strip date from end
      REGEXP_REPLACE( -- strip date from beginning
        METADATA$FILENAME,
        '^cloudassets/[\\d:/_]+',
        'cloudassets/'
      ),
      '([\\d:/_]+(S\\d+|A\\d+:\\d+).json|/[\\d]+)$'
    )
  ),
  log_time STRING AS TO_VARCHAR(TRY_TO_TIMESTAMP_NTZ(REGEXP_REPLACE(
    METADATA$FILENAME,
    '^.*/(\\d{4})/(\\d{2})/(\\d{2})/(\\d{2}):(\\d{2}):(\\d{2})[\\w\\d:/.]*$',
    '\\1-\\2-\\3T\\4:\\5:\\6Z'
  )))
~~~

seems like creating the dt and prefix definitions into `start_export_assets_job` might do the trick?

sfc-gh-dwyleczukstern: I addressed the prefix as requested and we didn't get any errors in the latest run

sfc-gh-pkommini:
lgtm after fix!

sfc-gh-pkommini (approved):
lgtm.

sfc-gh-afedorov (approved):

PR #515, opened 2021-03-24
[dc] Add SCC Findings to GCPIC

sfc-gh-afedorov:
runs without error and ingests rows as expected

### DIFF
--- a/src/connectors/utils.py
+++ b/src/connectors/utils.py
@@ -1,3 +1,4 @@
+from inspect import signature
 import random
 import multiprocessing as mp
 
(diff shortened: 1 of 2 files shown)


src/connectors/utils.py
```diff
@@ -139,3 +140,8 @@ def load_sqlfmt(file, params):
     # gen vars
```
sfc-gh-pkommini, line 140: Is this the doc string for the `appy_part` method? If so, it might be better to place the doc string inside the function.

sfc-gh-pkommini (approved):
lgtm.

src/connectors/utils.py
sfc-gh-afedorov: I think it's a stray comment I added awhile ago to pack.py that made its way into utils

sfc-gh-pkommini: I see. Might be okay to remove it. But also, maybe add a docstring as I wasn't able to fully grok what the `apply_part` is doing at first glance.

sfc-gh-pkommini (approved):
lgtm

PR #523, opened 2021-04-06
Adding CodeMirror 6

sfc-gh-nlele:
After making changes, the only part that was not working as supposed to was 'Format'. It was only updating the rule body with the expected formatting, but that wasn't getting reflected on the editor. To get around this, I tried to use a workaround, but now: when the formatter is used (with 'Format'), and then if I try to save it, a compilation error occurs. Apply/saving seems to be working fine if its done normally.


src/webui/frontend/src/routes/Dashboard/Dependencies.tsx
```diff
@@ -0,0 +1,75 @@
...
+    return (
+      <BasicLayout>
+        <Card
+          className={'card'}
+        
+        >
+            <div>
+              <LoadingOutlined /> Loading Rules...
+            </div>
+          ) : (
```
sfc-gh-afedorov, line 48: this seems malformed -- did you mean to include it in a separate branch, maybe?

sfc-gh-afedorov:
I'm not sure if I follow how the format button is working given that it's just toggling a bool -- maybe add a comment to explain? we'll likely refactor the Format functionality (or the server to be more accepting of alternative formatting), so no worries on keeping it working perfectly

src/webui/frontend/src/routes/Dashboard/Dependencies.tsx
sfc-gh-nlele: I had started working on a barebones data dependencies page, since this was a draft I didn't completely remove it yet, for working on it later. Will remove it now :)

sfc-gh-afedorov:
looks like the WhiteSource security checks are not working -- we'll need those before we can merge this change

PR #526, opened 2021-04-14
Add Jira list values in custom fields

sfc-gh-afedorov:
written very carefully and re-read twice


sfc-gh-dwyleczukstern (commented):
$ python
Python 3.8.5 (default, Sep 16 2020, 19:29:57)
[Clang 12.0.0 (clang-1200.0.31.1)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> field_value='[a,b,c,d]'
>>> [{'value': v} for v in field_value[1:-1].split(',')]
[{'value': 'a'}, {'value': 'b'}, {'value': 'c'}, {'value': 'd'}]
>>>

LGTM

PR #532, opened 2021-04-22
Modify https driver

sfc-gh-pkommini:
- Made changes to the https driver to add cursor capabilities for pagination
- Added two tests

### DIFF
--- a/infra/terraform/external_function/snowalert/lambda-code/drivers/process_https.py
+++ b/infra/terraform/external_function/snowalert/lambda-code/drivers/process_https.py
@@ -63,7 +63,7 @@ def process_row(
         if 'host' in req_auth and not req_host:
             req_host = req_auth['host']
 
-        if req_auth.get("host") != req_host:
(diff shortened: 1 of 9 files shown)


infra/terraform/external_function/snowalert/lambda-code/drivers/process_https.py
```diff
@@ -50,7 +51,7 @@ def process_row(
     }
     req_headers.setdefault('User-Agent', 'Snowflake SnowAlert External Function 1.0')
     if auth:
-        auth = decrypt_if_encrypted(auth)
+        auth = str(decrypt_if_encrypted(auth))
```
sfc-gh-afedorov, line 54: this shouldn't be necessary -- looking at [the docs](https://mypy.readthedocs.io/en/stable/common_issues.html) maybe mypy needs an explicit

    assert auth is not None

here?

sfc-gh-afedorov (approved):

PR #543, opened 2021-06-09
Add MPIM functionality to Slack handler

sfc-gh-rmahajan:
Case 1:  Single Email i.e Direct Message.
Case 2: More than 2 Emails i.e Group Message.
Case 3: Wrong Email for Direct Message.
Case 4: Wrong Email for Group Message.

### DIFF
--- a/src/runners/handlers/slack.py
+++ b/src/runners/handlers/slack.py
@@ -2,7 +2,6 @@ import os
 import json
 
 from slackclient import SlackClient
-from slack import WebClient
 
 from runners.helpers import log
 from runners.helpers import db
@@ -13,7 +12,7 @@ API_TOKEN = os.environ.get('SA_SLACK_API_TOKEN', os.environ.get('SLACK_API_TOKEN
 
 def message_template(vars):
     payload = None
-
+    # print("Hi")
     # remove handlers data, it might contain JSON incompatible strucutres
     vars['alert'].pop('HANDLERS')
 
@@ -68,7 +67,6 @@ def handle(
     slack_token = vault.decrypt_if_encrypted(slack_token_ct)
 
     sc = SlackClient(slack_token)
-    client = WebClient(slack_token)
     # otherwise we will retrieve email from assignee and use it to identify Slack user
     # Slack user id will be assigned as a channel
 
@@ -76,15 +74,24 @@ def handle(
 
     if recipient_email is not None:
         if isinstance(recipient_email, str):
-            result = sc.api_call("users.lookupByEmail", email=recipient_email)
+                user_object = sc.api_call("users.lookupByEmail", email=recipient_email)
+                user = user_object['user']
+                users.append(user['id'])
+                result = sc.api_call("conversations.open", users=users)
+
 
         else:
-            result = client.conversations_open(users= recipient_email)
+            users = []
+            for email in recipient_email:
+                user_object = sc.api_call("users.lookupByEmail", email=email)
+                user = user_object['user']
+                users.append(user['id'])
+            result = sc.api_call("conversations.open", users=users)
 
         # log.info(f'Slack user info for {email}', result)
 
         if result['ok'] is True and 'error' not in result:
-            user = result['user']
+            user = result['channel']
             userid = user['id']
         else:
             log.error(f'Cannot identify  Slack user for email {recipient_email}')
--- a/src/runners/handlers/slack.py
+++ b/src/runners/handlers/slack.py
@@ -12,6 +12,7 @@ API_TOKEN = os.environ.get('SA_SLACK_API_TOKEN', os.environ.get('SLACK_API_TOKEN
 
 def message_template(vars):
     payload = None
+
     # remove handlers data, it might contain JSON incompatible strucutres
     vars['alert'].pop('HANDLERS')
 
@@ -64,8 +65,8 @@ def handle(
 ):
     slack_token_ct = slack_api_token or api_token
     slack_token = vault.decrypt_if_encrypted(slack_token_ct)
-
     sc = SlackClient(slack_token)
+
     # otherwise we will retrieve email from assignee and use it to identify Slack user
     # Slack user id will be assigned as a channel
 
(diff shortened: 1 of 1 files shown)


src/runners/handlers/slack.py
```diff
@@ -67,14 +68,18 @@ def handle(
     slack_token = vault.decrypt_if_encrypted(slack_token_ct)
 
     sc = SlackClient(slack_token)
-
+    client = WebClient(slack_token)
```
sfc-gh-afedorov, line 71: could re-use the SlackClient we already have for consistency? I think it can make any API call via `sc.api_call` as used below

src/runners/handlers/slack.py
```diff
@@ -67,14 +68,18 @@ def handle(
...
 
     title = alert['TITLE']
 
     if recipient_email is not None:
-        result = sc.api_call("users.lookupByEmail", email=recipient_email)
+        if isinstance(recipient_email, str):
+            result = sc.api_call("users.lookupByEmail", email=recipient_email)
+
+        else:
+            result = client.conversations_open(users= recipient_email)
```
sfc-gh-afedorov, line 82: looking at [the docs](https://api.slack.com/methods/conversations.open), it seems "conversation.open" takes a list of user id's, not user emails, so first this code needs to resolve a list of user emails to a list of user id's here

then below, you need to resolve swap out `sc.api_call("chat.postMessage", ...)` with `sc.api_call("conversations.open", ...)`

sfc-gh-afedorov (changes requested):
this is a good first attempt, but I don't think this matches the API. would it help to get an invite to our Sandbox Slack to test out the API?

src/runners/handlers/slack.py
```diff
@@ -67,19 +66,31 @@ def handle(
...
 
     sc = SlackClient(slack_token)
-
     # otherwise we will retrieve email from assignee and use it to identify Slack user
     # Slack user id will be assigned as a channel
 
     title = alert['TITLE']
 
     if recipient_email is not None:
-        result = sc.api_call("users.lookupByEmail", email=recipient_email)
```
sfc-gh-afedorov, line 77: it seems result *used* to be the response to `users.lookupByEmail` but now it is the response to `conversations.open` -- could you read the rest of the function to make sure it makes sense, still, and update e.g. the `log.error` on line 96 with the new meaning?

src/runners/handlers/slack.py
```diff
@@ -67,19 +66,31 @@ def handle(
...
+        if isinstance(recipient_email, str):
+                user_object = sc.api_call("users.lookupByEmail", email=recipient_email)
```
sfc-gh-afedorov, line 76: I can do this before merging, but would save us an extra step if you could install Python Black in your text editor and fix e.g. this indentation to be 4 spaces.

src/runners/handlers/slack.py
```diff
@@ -67,19 +66,31 @@ def handle(
...
+                user = user_object['user']
+                users.append(user['id'])
+                result = sc.api_call("conversations.open", users=users)
+
+
+        else:
+            users = []
+            for email in recipient_email:
+                user_object = sc.api_call("users.lookupByEmail", email=email)
+                user = user_object['user']
```
sfc-gh-afedorov, line 86: both here and above, when a user is not found, we will likely throw an exception without details about which user it err'd on. it might help error a little more gracefully via —

~~~py
assert 'user' in user_object, f'error while finding user "{email}"'
~~~

sfc-gh-afedorov (commented):
nice! a few nitpicks / error cases aside, this looks to my eye like it should work, but please do handle foreseeable errors and update the description when you've tested it.

sfc-gh-rmahajan:
@sfc-gh-afedorov Changes are done.

sfc-gh-afedorov:
> @sfc-gh-afedorov Changes are done.

please include testing method and evidence in the PR description

sfc-gh-rmahajan:
@sfc-gh-afedorov please review this.

src/runners/handlers/slack.py
```diff
@@ -163,10 +178,9 @@ def handle(
...
         if response['ok'] is False:
-            log.error(f"Slack handler error", response['error'])
-            return None
+            raise RuntimeError(f"Slack handler error {response['error']}")
 
         if 'message' in response:
             del response['message']
 
-    return response
+    return response
```
sfc-gh-afedorov, line 186: don't strip trailing newlines

src/runners/handlers/slack.py
```diff
@@ -74,30 +73,46 @@ def handle(
...
-            channel = userid
+            channel = user_id
             log.info(
                 f'Creating new SLACK message for {title} for user {recipient_email}'
             )
         else:
             log.error(f'Cannot identify assignee email')
             return None
-
+
```
sfc-gh-afedorov, line 115: don't add vestigial indents

src/runners/handlers/slack.py
```diff
@@ -74,30 +73,46 @@ def handle(
...
+                        
+        elif(recipient_email, list):
+            users = []
+            for email in recipient_email:
+                user = sc.api_call("users.lookupByEmail", email=email)
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {email}')
+                    return None
+                users.append(user['user']['id'])
+                user_ids = ",".join(users)
```
sfc-gh-afedorov, line 98: seems likely you want to do this once after the for loop, not on every iteration

src/runners/handlers/slack.py
```diff
@@ -74,30 +73,46 @@ def handle(
...
+            for email in recipient_email:
+                user = sc.api_call("users.lookupByEmail", email=email)
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {email}')
+                    return None
+                users.append(user['user']['id'])
+                user_ids = ",".join(users)                  
+                #converting list to comma seperated string
+            result = sc.api_call("conversations.open", users=user_ids)
+            user_id = result['channel']['id']
```
sfc-gh-afedorov, line 101: let's call this `channel_id` so it's less confusing

src/runners/handlers/slack.py
```diff
@@ -74,30 +73,46 @@ def handle(
...
+        if isinstance(recipient_email, str):
+                user = sc.api_call("users.lookupByEmail", email=recipient_email)
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {recipient_email}')
+                    return None
+                    
+                else:
+                    userid = user['user']['id']
+                    result = sc.api_call("conversations.open", users=userid)
+                    user_id = result['channel']['id']
```
sfc-gh-afedorov, line 85: what's the difference between an IM to one user and a channel with only one other participant?

src/runners/handlers/slack.py
```diff
@@ -74,30 +73,46 @@ def handle(
...
+                        log.error(f'Error ocurred while opening conversation channel')
+                        return None
+                        
+        elif(recipient_email, list):
+            users = []
+            for email in recipient_email:
+                user = sc.api_call("users.lookupByEmail", email=email)
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {email}')
+                    return None
```
sfc-gh-afedorov, line 96: I think you might want to `continue` rather than `return` here, e.g. if one of the users has changed their names or left the company, no need to throw errors if there are other users to DM

src/runners/handlers/slack.py
```diff
@@ -74,30 +73,46 @@ def handle(
...
+                user = sc.api_call("users.lookupByEmail", email=recipient_email)
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {recipient_email}')
+                    return None
+                    
+                else:
+                    userid = user['user']['id']
+                    result = sc.api_call("conversations.open", users=userid)
+                    user_id = result['channel']['id']
+                    if not result['ok']:
```
sfc-gh-afedorov, line 86: you likely want to test this *before* referencing `result['channel']['id']`

src/runners/handlers/slack.py
```diff
@@ -74,30 +73,46 @@ def handle(
...
+                        log.error(f'Error ocurred while opening conversation channel')
+                        return None
+                        
+        elif(recipient_email, list):
```
sfc-gh-afedorov, line 90: I think you meant to use `isinstance`

src/runners/handlers/slack.py
```diff
@@ -74,30 +73,46 @@ def handle(
...
+            users = []
+            for email in recipient_email:
+                user = sc.api_call("users.lookupByEmail", email=email)
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {email}')
+                    return None
+                users.append(user['user']['id'])
+                user_ids = ",".join(users)                  
+                #converting list to comma seperated string
+            result = sc.api_call("conversations.open", users=user_ids)
```
sfc-gh-afedorov, line 100: `if not result['ok']` might help here, too

src/runners/handlers/slack.py
```diff
@@ -74,30 +73,46 @@ def handle(
...
+            user_id = result['channel']['id']
+            
     # check if channel exists, if yes notification will be delivered to the channel
     if channel is not None:
         log.info(f'Creating new SLACK message for {title} in channel', channel)
     else:
         if recipient_email is not None:
-            channel = userid
+            channel = user_id
```
sfc-gh-afedorov, line 108: I'm not sure what this rename serves, but if you rename it channel_id above, it seems to make sense to keep that name throughout.

src/runners/handlers/slack.py
sfc-gh-rmahajan: Sorry, didn't get this.

src/runners/handlers/slack.py
sfc-gh-rmahajan: If only one user, we are sending it mostly as string from object construct .sql file. So, user can send email as a single string or a list(will work in case of single email also).

src/runners/handlers/slack.py
sfc-gh-afedorov: there is a newline at the end of the file that you took out in this PR -- please put it back

sfc-gh-afedorov:
re-submitted as #544

PR #544, opened 2021-06-24
Slack Group Message

sfc-gh-rmahajan:
Slack Group Messages

case 1: When single recipient_email is passed i.e direct message:
Screenshot 2021-06-25 at 1 02 41 PM
Output:
Screenshot 2021-06-25 at 1 11 59 PM

Explaination:
Here, for Single Email, string will be be passed as recipient so, convert it into list.

if isinstance(recipient_email, str):
   recipient_email = [recipient_email]

case 2: When more than one recipient_email is passed i.e group slack creation:
Screenshot 2021-06-25 at 1 08 42 PM
Output:
Screenshot 2021-06-25 at 1 09 09 PM

Explaination:
If we get a list of emails, we need to iterate and get userId for each email and store it in as a list and then send all the ids as string by calling an API.

      for email in recipient_email:
          response = sc.api_call("users.lookupByEmail", email=email)
          if not response['ok']:
              log.error(f'Cannot identify Slack user for email {email}')
              continue
          users.append(response['user']['id'])
      user_ids = ",".join(users)                  
      result = sc.api_call("conversations.open", users=user_ids)

### DIFF
--- a/src/runners/handlers/slack.py
+++ b/src/runners/handlers/slack.py
@@ -70,42 +70,25 @@ def handle(
     # otherwise we will retrieve email from assignee and use it to identify Slack user
     # Slack user id will be assigned as a channel
 
-    # title = alert['TITLE']
-    title = 'title'
+    title = alert['TITLE']
+    
     if recipient_email is not None:
         if isinstance(recipient_email, str):
-                user = sc.api_call("users.lookupByEmail", email=recipient_email)
-                print(user)
-                if not user['ok']:
-                    log.error(f'Cannot identify Slack user for email {recipient_email}')
-                    return None
-                    
-                else:
-                    userid = user['user']['id']
-                    result = sc.api_call("conversations.open", users=userid)
-                    if result['ok']:
-                        channel_id = result['channel']['id']
-                    else:
-                        log.error(f'Error ocurred while opening conversation channel')
-                        return None
+            recipient_email = [recipient_email]
                         
-        elif isinstance(recipient_email, list):
-            users = []
-            for email in recipient_email:
-                user = sc.api_call("users.lookupByEmail", email=email)
-                print(user)
-                if not user['ok']:
-                    log.error(f'Cannot identify Slack user for email {email}')
-                    continue
-                users.append(user['user']['id'])
-            user_ids = ",".join(users)                  
-                #converting list to comma seperated string
-            result = sc.api_call("conversations.open", users=user_ids)
-            if result['ok']:
-                channel_id = result['channel']['id']
-            else:
-                log.error(f'Error ocurred while opening conversation channel')
-                return None
+        users = []
+        for email in recipient_email:
+            response = sc.api_call("users.lookupByEmail", email=email)
+            if not response['ok']:
+                log.error(f'Cannot identify Slack user for email {email}')
+                continue
+            users.append(response['user']['id'])
+        user_ids = ",".join(users)                  
+        result = sc.api_call("conversations.open", users=user_ids)
+        if result['ok']:
+            channel_id = result['channel']['id']
+        else:
+            raise RuntimeError(f"Error ocurred while opening conversation channel")
             
     # check if channel exists, if yes notification will be delivered to the channel
     if channel is not None:
@@ -117,8 +100,7 @@ def handle(
                 f'Creating new SLACK message for {title} for user {recipient_email}'
             )
         else:
-            log.error(f'Cannot identify assignee email')
-            return None
+            raise RuntimeError("missing both 'channel' and 'recipient_email' param")
     
     text = title
 
--- a/src/runners/handlers/slack.py
+++ b/src/runners/handlers/slack.py
@@ -71,11 +71,11 @@ def handle(
     # Slack user id will be assigned as a channel
 
     title = alert['TITLE']
-    
+
     if recipient_email is not None:
         if isinstance(recipient_email, str):
             recipient_email = [recipient_email]
-                        
+
         users = []
         for email in recipient_email:
             response = sc.api_call("users.lookupByEmail", email=email)
@@ -83,13 +83,13 @@ def handle(
                 log.error(f'Cannot identify Slack user for email {email}')
                 continue
             users.append(response['user']['id'])
-        user_ids = ",".join(users)                  
+        user_ids = ",".join(users)
         result = sc.api_call("conversations.open", users=user_ids)
         if result['ok']:
             channel_id = result['channel']['id']
         else:
             raise RuntimeError(f"Error ocurred while opening conversation channel")
-            
+
     # check if channel exists, if yes notification will be delivered to the channel
     if channel is not None:
         log.info(f'Creating new SLACK message for {title} in channel', channel)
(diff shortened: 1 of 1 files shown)


src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
-
     sc = SlackClient(slack_token)
 
     # otherwise we will retrieve email from assignee and use it to identify Slack user
     # Slack user id will be assigned as a channel
 
-    title = alert['TITLE']
-
+    # title = alert['TITLE']
+    title = 'title'
```
sfc-gh-afedorov, line 74: this change seem to be left-over from testing

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
-            user = result['user']
-            userid = user['id']
-        else:
-            raise RuntimeError(f'no Slack user for email {recipient_email}')
-
+        if isinstance(recipient_email, str):
+                user = sc.api_call("users.lookupByEmail", email=recipient_email)
+                print(user)
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {recipient_email}')
```
sfc-gh-afedorov, line 80: if there's an error, you want to raise it rather than returning None (which signals everything was Ok)

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
-
-        if result['ok'] is True and 'error' not in result:
-            user = result['user']
-            userid = user['id']
-        else:
-            raise RuntimeError(f'no Slack user for email {recipient_email}')
-
+        if isinstance(recipient_email, str):
+                user = sc.api_call("users.lookupByEmail", email=recipient_email)
+                print(user)
```
sfc-gh-afedorov, line 78: remove extra print

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
-        # log.info(f'Slack user info for {email}', result)
-
-        if result['ok'] is True and 'error' not in result:
-            user = result['user']
-            userid = user['id']
-        else:
-            raise RuntimeError(f'no Slack user for email {recipient_email}')
-
+        if isinstance(recipient_email, str):
+                user = sc.api_call("users.lookupByEmail", email=recipient_email)
```
sfc-gh-afedorov, line 77: this is better named "response"

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
+                print(user)
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {recipient_email}')
+                    return None
+                    
+                else:
+                    userid = user['user']['id']
+                    result = sc.api_call("conversations.open", users=userid)
```
sfc-gh-afedorov, line 85: this is better called "response", since it includes both the result and the status code

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
+                    if result['ok']:
+                        channel_id = result['channel']['id']
+                    else:
+                        log.error(f'Error ocurred while opening conversation channel')
```
sfc-gh-afedorov, line 89: if you can't continue (as here), raise exception rather than logging error

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
-
+        if isinstance(recipient_email, str):
+                user = sc.api_call("users.lookupByEmail", email=recipient_email)
+                print(user)
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {recipient_email}')
+                    return None
+                    
+                else:
+                    userid = user['user']['id']
```
sfc-gh-afedorov, line 84: do you intend to use `userid` more than once? it's better to skip it if you don't and inline `response['user']['id']` as a parameter in the next line

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
+                        channel_id = result['channel']['id']
+                    else:
+                        log.error(f'Error ocurred while opening conversation channel')
+                        return None
+                        
+        elif isinstance(recipient_email, list):
+            users = []
+            for email in recipient_email:
+                user = sc.api_call("users.lookupByEmail", email=email)
+                print(user)
```
sfc-gh-afedorov, line 96: remove vestigial print statements. they may seem harmless, but the can actually increase the sensitivity classification of logs, which ideally should not contain employee PII. if you raise or return this value, the error will be stored in Snowflake.

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {email}')
```
sfc-gh-afedorov, line 98: raise exception instead of logging

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
+                    continue
+                users.append(user['user']['id'])
+            user_ids = ",".join(users)                  
+                #converting list to comma seperated string
+            result = sc.api_call("conversations.open", users=user_ids)
+            if result['ok']:
+                channel_id = result['channel']['id']
+            else:
+                log.error(f'Error ocurred while opening conversation channel')
```
sfc-gh-afedorov, line 107: raise instead of returning

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
+            users = []
+            for email in recipient_email:
+                user = sc.api_call("users.lookupByEmail", email=email)
+                print(user)
+                if not user['ok']:
+                    log.error(f'Cannot identify Slack user for email {email}')
+                    continue
+                users.append(user['user']['id'])
+            user_ids = ",".join(users)                  
+                #converting list to comma seperated string
```
sfc-gh-afedorov, line 102: no need for this comment -- everyone should know what `str.join` does

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
+                log.error(f'Error ocurred while opening conversation channel')
+                return None
+            
     # check if channel exists, if yes notification will be delivered to the channel
     if channel is not None:
         log.info(f'Creating new SLACK message for {title} in channel', channel)
     else:
         if recipient_email is not None:
-            channel = userid
+            channel = channel_id
```
sfc-gh-afedorov, line 115: no need to rename it, and adds unnecessary complexity

src/runners/handlers/slack.py
```diff
@@ -65,37 +65,61 @@ def handle(
...
             log.info(
                 f'Creating new SLACK message for {title} for user {recipient_email}'
             )
         else:
-            raise RuntimeError("missing both 'channel' and 'recipient_email' param")
-
+            log.error(f'Cannot identify assignee email')
```
sfc-gh-afedorov, line 120: this was better as a `raise`, since that will record that the handler did not complete what it was intended to compelte

sfc-gh-afedorov (changes requested):
there's still a couple of fixes left, but you are getting a lot closer

sfc-gh-afedorov (approved):
this looks good, thank you for your patience :)

PR #545, opened 2021-06-28
Slack group messages

sfc-gh-rmahajan:
Changes for single Message.


src/runners/handlers/slack.py
```diff
@@ -65,31 +65,48 @@ def handle(
...
     slack_token = vault.decrypt_if_encrypted(slack_token_ct)
-
     sc = SlackClient(slack_token)
 
     # otherwise we will retrieve email from assignee and use it to identify Slack user
     # Slack user id will be assigned as a channel
 
-    title = alert['TITLE']
-
+    # title = alert['TITLE']
```
sfc-gh-afedorov, line 73: I'm not sure why you keep commenting this out when you can just add it to the test? another solution might be to use `alert.get('title', '')` if you absolutely must test it with an empty dict

sfc-gh-afedorov (changes requested):
please add how this was tested to the PR description

sfc-gh-afedorov (approved):

PR #546, opened 2021-06-29
Add azure collect network interfaces

sfc-gh-dwyleczukstern:
We need to be able to collect network interfaces to tie VMs to flow logs
https://docs.microsoft.com/en-us/rest/api/virtualnetwork/network-interfaces/list-all#code-try-0


sfc-gh-afedorov:
already have it

PR #550, opened 2021-08-16
[Fix] Collecting IAM managed/inline policy data for users/groups/roles

stevenliu-czi:
This has not been tested and was written based off pattern matching on previous examples within the files.

### DIFF
--- a/migrations/v1_9_6-v1_10_0.md
+++ b/migrations/v1_9_6-v1_10_0.md
@@ -267,7 +267,7 @@ CREATE OR REPLACE TABLE data.aws_collect_iam_get_user_policy (
     policy_name STRING,
     user_name STRING,
     error VARIANT,
-    policy_document STRING,
+    policy_document STRING
 )
 ;
 GRANT SELECT, INSERT
@@ -283,7 +283,7 @@ CREATE OR REPLACE TABLE data.aws_collect_iam_list_group_policies (
     account_id STRING,
     group_name STRING,
     error VARIANT,
-    policy_name STRING,
+    policy_name STRING
 )
 ;
 GRANT SELECT, INSERT
@@ -300,7 +300,7 @@ CREATE OR REPLACE TABLE data.aws_collect_iam_get_group_policy (
     policy_name STRING,
     group_name STRING,
     error VARIANT,
-    policy_document STRING,
+    policy_document STRING
 )
 ;
 GRANT SELECT, INSERT
@@ -317,7 +317,7 @@ CREATE OR REPLACE TABLE data.aws_collect_iam_list_attached_role_policies (
     role_name STRING,
     error VARIANT,
     policy_name STRING,
-    policy_arn STRING,
+    policy_arn STRING
 )
 ;
 GRANT SELECT, INSERT
--- a/migrations/v1_9_6-v1_10_0.md
+++ b/migrations/v1_9_6-v1_10_0.md
@@ -258,3 +258,70 @@ ON data.aws_collect_ec2_describe_snapshots
 TO ROLE snowalert
 ;
 ```
+
+## Get AWS User Inline Policy Documents
+```sql
+CREATE OR REPLACE TABLE data.aws_collect_iam_get_user_policy (
+    recorded_at TIMESTAMP_LTZ,
+    account_id STRING,
+    policy_name STRING,
+    user_name STRING,
+    error VARIANT,
+    policy_document STRING,
+)
+;
+GRANT SELECT, INSERT
+ON data.aws_collect_iam_get_user_policy
+TO ROLE snowalert
+;
+```
+
+## Get AWS Group Inline Policy List
+```sql
+CREATE OR REPLACE TABLE data.aws_collect_iam_list_group_policies (
+    recorded_at TIMESTAMP_LTZ,
+    account_id STRING,
+    group_name STRING,
+    error VARIANT,
+    policy_name STRING,
+)
+;
+GRANT SELECT, INSERT
+ON data.aws_collect_iam_list_group_policies
+TO ROLE snowalert
+;
+```
+
+## Get AWS Group Inline Policy Documents
+```sql
+CREATE OR REPLACE TABLE data.aws_collect_iam_get_group_policy (
+    recorded_at TIMESTAMP_LTZ,
+    account_id STRING,
+    policy_name STRING,
+    group_name STRING,
+    error VARIANT,
+    policy_document STRING,
+)
+;
+GRANT SELECT, INSERT
+ON data.aws_collect_iam_get_group_policy
+TO ROLE snowalert
+;
+```
+
+## Get AWS Role Managed Policies
+```sql
+CREATE OR REPLACE TABLE data.aws_collect_iam_list_attached_role_policies (
+    recorded_at TIMESTAMP_LTZ,
+    account_id STRING,
+    role_name STRING,
+    error VARIANT,
+    policy_name STRING,
+    policy_arn STRING,
+)
+;
+GRANT SELECT, INSERT
+ON data.aws_collect_iam_list_attached_role_policies
+TO ROLE snowalert
+;
+```
--- a/src/connectors/aws_collect.py
+++ b/src/connectors/aws_collect.py
@@ -1030,10 +1030,10 @@ API_METHOD_SPECS: Dict[str, dict] = {
         },
         'children': [
             {
-                'methods': {
+                'methods': [
                     'iam.list_attached_group_policies',
                     'iam.list_group_policies',
-                }
+                ],
                 'args': {'GroupName': 'group_name'},
             }
         ],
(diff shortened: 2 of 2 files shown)


migrations/v1_9_6-v1_10_0.md
```diff
@@ -258,3 +258,70 @@ ON data.aws_collect_ec2_describe_snapshots
...
+
+## Get AWS User Inline Policy Documents
+```sql
+CREATE OR REPLACE TABLE data.aws_collect_iam_get_user_policy (
+    recorded_at TIMESTAMP_LTZ,
+    account_id STRING,
+    policy_name STRING,
+    user_name STRING,
+    error VARIANT,
+    policy_document STRING,
```
sfc-gh-afedorov, line 270: trailing commas won't work in Snowflake

src/connectors/aws_collect.py
```diff
@@ -391,6 +391,15 @@
...
         ('policy_name', 'STRING'),
     ],
+    # https://docs.aws.amazon.com/cli/latest/reference/iam/get-user-policy.html#output
+    'iam_get_user_policy': [
+        ('recorded_at', 'TIMESTAMP_LTZ'),
+        ('account_id', 'STRING'),
+        ('user_name', 'STRING'),
+        ('policy_name', 'STRING'),
+        ('error', 'VARIANT'),
+        ('policy_document', 'STRING'),
```
sfc-gh-afedorov, line 401: I think we can make these VARIANT, as well, and parse JSON or YAML based on what's in the value?

sfc-gh-afedorov (changes requested):
lgtm with a few minor suggestions —

1. remove trailing commas in SQL where it wouldn't work
2. if possible, parse the policies into VARIANT's instead of leaving them as STRING's

also, please add description of how this was tested -- it's Ok if it wasn't, we can help test, but no need to repeat if you've already done so

src/connectors/aws_collect.py
stevenliu-czi: Gotcha, would we want to reflect the changes to VARIANT in both aws_collect.py AND the migration md?

sfc-gh-afedorov: Yes, and also the API schema, which should declare it as parsed JSON akin to [here](https://github.com/snowflakedb/SnowAlert/blob/77127291bd8bfde15a8d5d46ce8ad0fbf276faed/src/connectors/aws_collect.py#L1305). The [docs](https://docs.aws.amazon.com/cli/latest/reference/iam/get-user-policy.html#output) are a little unclear as to whether this string might also sometimes be YAML if deployed by CloudFormation, so testing would make sense. I can run it against our infra but we don't use CloudFormation much. Do you have a test-case where such a policy was deployed by CF or should we set up a separate test run?

stevenliu-czi: Our policies are created through terraform, so I don't believe we have a test-case for CF

stevenliu-czi: Is there an example I could follow for which fields in the API scheme should be VARIANT? I'm just a bit unsure as it seems like even the '[s3_get_bucket_policy](https://github.com/snowflakedb/SnowAlert/blob/77127291bd8bfde15a8d5d46ce8ad0fbf276faed/src/connectors/aws_collect.py#L568-L573)' has its fields declared with the same types as mine (except with the extra json parsed policy)

sfc-gh-afedorov: Any of the fields which are documented as always returning JSON should be stored in a VARIANT. You can also check if the example output is JSON. It wasn't part of the original feature set and we want to be backwards compatible so there's both `policy` and `policy_json_parsed` columns that get set by the code that handles the ParsedCol case [here](https://github.com/snowflakedb/SnowAlert/blob/77127291bd8bfde15a8d5d46ce8ad0fbf276faed/src/connectors/aws_collect.py#L1620-L1623).

There's also `content_csv_parsed` in `iam_get_credential_report` [here](https://github.com/snowflakedb/SnowAlert/blob/77127291bd8bfde15a8d5d46ce8ad0fbf276faed/src/connectors/aws_collect.py#L1014) which is what drove this use-case. YAML would be similar to CSV in that Snowflake does not have native parsing support and we would prefer to pre-process these before landing them in Snowflake.

stevenliu-czi:
None of the commits have been tested and the new code changes were written based off pattern matching on previous examples within the files.

stevenliu-czi:
[Updated comment for new commit] Added syntax fixes. Still not tested.

sfc-gh-afedorov:
tested on one of our internal accounts

~~~
(ve) {prod}[AAA] src % SA_ENV=dev ipython --pdb -- runners/connectors_runner.py AWS_COLLECT_ORGANIZATIONS_LIST_ACCOUNTS_CONNECTION --run-now --dryrun False --debug --org-account-ids 'CCC' --collect-apis 'iam.list_groups,iam.list_users,iam.list_roles'
/usr/local/lib/python3.8/site-packages/IPython/core/interactiveshell.py:936: UserWarning: Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv.
  warn("Attempting to work in a virtualenv. If you encounter problems, please "
[57339] -- START DC AWS_COLLECT_ORGANIZATIONS_LIST_ACCOUNTS_CONNECTION --
[57339] RUN metadata recorded.
[57339] botocore.errorfactory.AccessDeniedException: An error occurred (AccessDeniedException) when calling the ListAccounts operation: You don't have permissions to access this resource.
[57339] inserting 1 values into data.AWS_COLLECT_ORGANIZATIONS_LIST_ACCOUNTS_CONNECTION
[57339] progress: starting 3, queued 0
[57339] inserting 2 values into data.aws_collect_iam_list_groups
[57339] finished iam.list_groups {'number of rows inserted': 2}
[57339] inserting 23 values into data.aws_collect_iam_list_users
[57339] finished iam.list_users {'number of rows inserted': 23}
[57339] inserting 134 values into data.aws_collect_iam_list_roles
[57339] finished iam.list_roles {'number of rows inserted': 134}
[57339] progress: starting 410, queued 0
[57339] botocore.errorfactory.NoSuchEntityException: An error occurred (NoSuchEntity) when calling the GetLoginProfile operation: Login Profile for User BBB cannot be found.
[57339] botocore.errorfactory.NoSuchEntityException: An error occurred (NoSuchEntity) when calling the GetLoginProfile operation: Login Profile for User BBB cannot be found.
[57339] botocore.errorfactory.NoSuchEntityException: An error occurred (NoSuchEntity) when calling the GetLoginProfile operation: Login Profile for User BBB cannot be found.
[57339] botocore.errorfactory.NoSuchEntityException: An error occurred (NoSuchEntity) when calling the GetLoginProfile operation: Login Profile for User BBB cannot be found.
[57339] botocore.errorfactory.NoSuchEntityException: An error occurred (NoSuchEntity) when calling the GetLoginProfile operation: Login Profile for User BBB cannot be found.
[57339] botocore.errorfactory.NoSuchEntityException: An error occurred (NoSuchEntity) when calling the GetLoginProfile operation: Login Profile for User BBB cannot be found.
[57339] inserting 2 values into data.aws_collect_iam_list_group_policies
[57339] finished iam.list_group_policies {'number of rows inserted': 2}
[57339] inserting 3 values into data.aws_collect_iam_list_attached_group_policies
[57339] finished iam.list_attached_group_policies {'number of rows inserted': 3}
[57339] inserting 25 values into data.aws_collect_iam_list_access_keys
[57339] finished iam.list_access_keys {'number of rows inserted': 25}
[57339] inserting 23 values into data.aws_collect_iam_list_mfa_devices
[57339] finished iam.list_mfa_devices {'number of rows inserted': 23}
[57339] inserting 23 values into data.aws_collect_iam_list_user_policies
[57339] finished iam.list_user_policies {'number of rows inserted': 23}
[57339] inserting 28 values into data.aws_collect_iam_list_groups_for_user
[57339] finished iam.list_groups_for_user {'number of rows inserted': 28}
[57339] inserting 24 values into data.aws_collect_iam_list_attached_user_policies
[57339] finished iam.list_attached_user_policies {'number of rows inserted': 24}
[57339] inserting 23 values into data.aws_collect_iam_get_login_profile
[57339] finished iam.get_login_profile {'number of rows inserted': 23}
[57339] inserting 174 values into data.aws_collect_iam_list_attached_role_policies
[57339] finished iam.list_attached_role_policies {'number of rows inserted': 174}
[57339] inserting 140 values into data.aws_collect_iam_list_role_policies
[57339] finished iam.list_role_policies {'number of rows inserted': 140}
[57339] progress: starting 52, queued 0
[57339] inserting 49 values into data.aws_collect_iam_get_role_policy
[57339] finished iam.get_role_policy {'number of rows inserted': 49}
[57339] inserting 2 values into data.aws_collect_iam_get_user_policy
[57339] finished iam.get_user_policy {'number of rows inserted': 2}
[57339] inserting 1 values into data.aws_collect_iam_get_group_policy
[57339] finished iam.get_group_policy {'number of rows inserted': 1}
[57339] RUN metadata recorded.
[57339] -- END DC AWS_COLLECT_ORGANIZATIONS_LIST_ACCOUNTS_CONNECTION --
SA_ENV=dev ipython --pdb -- runners/connectors_runner.py  --run-now --dryrun   11.39s user 1.07s system 24% cpu 50.431 total
~~~

sfc-gh-afedorov (approved):
lgtm

PR #567, opened 2021-11-22
[awsic] fix oids param, options

sfc-gh-afedorov:
when passed in the command line, parameters were automatically parsed into a list of integers. instead, we want to treat it as a list of strings.


sfc-gh-pkommini (approved):
lgtm

PR #568, opened 2021-11-29
Update Web UI dependencies

sfc-gh-nlele:


sfc-gh-afedorov:
let's start it simple --

1. run a `yarn upgrade-interactive` and update everything except TypeScript. commit all those changes into the first commit titled "Upgrade WebUI dependencies".
2. run `yarn upgrade-interactive` and update TypeScript, fixing any code that requires changes, and commit it separately titled "Upgrade WebUI TypeScript"

PR #569, opened 2021-11-30
Update WebUI dependencies

sfc-gh-nlele:
Installed dependencies, built and ran the WebUI to confirm that there are no breaking changes.

### DIFF
--- a/src/webui/frontend/package.json
+++ b/src/webui/frontend/package.json
@@ -51,7 +51,7 @@
   },
   "proxy": "http://localhost:8000",
   "devDependencies": {
-    "@types/classnames": "2.3.1",
+    "@types/classnames": "^2.3.1",
     "@types/enquire.js": "^2.1.1",
     "@types/history": "^4.7.8",
     "@types/jwt-decode": "^2.2.1",
--- a/src/webui/frontend/package.json
+++ b/src/webui/frontend/package.json
@@ -51,7 +51,7 @@
   },
   "proxy": "http://localhost:8000",
   "devDependencies": {
-    "@types/classnames": "^2.2.10",
+    "@types/classnames": "2.3.1",
     "@types/enquire.js": "^2.1.1",
     "@types/history": "^4.7.8",
     "@types/jwt-decode": "^2.2.1",
--- a/src/webui/frontend/package.json
+++ b/src/webui/frontend/package.json
@@ -10,7 +10,7 @@
     "@types/react": "^16.9.49",
     "@types/react-router": "^5.1.4",
     "antd": "^4.6.5",
-    "axios": "^0.20.0",
+    "axios": "^0.21.2",
     "classnames": "^2.2.6",
     "connected-react-router": "^6.8.0",
     "enquire.js": "^2.1.6",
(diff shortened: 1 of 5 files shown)


src/webui/frontend/src/actions/rules.ts
```diff
@@ -35,7 +35,9 @@ export const loadSnowAlertRules = () => async (dispatch: Dispatch, getState: Get
       const response = await api.loadSnowAlertRules();
       dispatch(LoadRulesActions.loadSnowAlertRulesSuccess(response.rules));
     } catch (error) {
-      dispatch(LoadRulesActions.loadSnowAlertRulesFailure(error.message));
+      if (error instanceof Error) {
```
sfc-gh-afedorov, line 38: does it work and might it make more sense to type hint `: any` here, as well? has this been tested and will this ever not exec a dispatch?

src/webui/frontend/src/actions/auth.ts
```diff
@@ -27,17 +27,19 @@ export const LoginActions = {
 
 export type LoginActions = ActionsUnion<typeof LoginActions>;
 
-export const oauthRedirect = (account: string, role: string, database: string, warehouse: string, returnHref: string) => async (dispatch: Dispatch) => {
-  try {
-    const response = await api.oauthRedirect({account, role, database, warehouse, returnHref});
-    if (response.url) {
-      console.log('navigating', response.url);
-      navigate(response.url);
+export const oauthRedirect =
```
sfc-gh-afedorov, line 30: just reformatting, no change to code here, right?

src/webui/frontend/package.json
```diff
@@ -51,7 +51,7 @@
   },
   "proxy": "http://localhost:8000",
   "devDependencies": {
-    "@types/classnames": "^2.2.10",
+    "@types/classnames": "2.3.1",
```
sfc-gh-afedorov, line 54: did you mean to remove the `^` here?

sfc-gh-afedorov (approved):
lgtm, please add how it was tested before merging.

src/webui/frontend/src/actions/rules.ts
sfc-gh-nlele: It will not exec a dispatch in case something odd, which is not an error is thrown, which may make `error.message` result in an `undefined`.  Sorry yes, `: any` seems like a better choice here, since it would accommodate cases like that too.

src/webui/frontend/src/actions/auth.ts
sfc-gh-nlele: Yes, no code changes here.

src/webui/frontend/package.json
sfc-gh-nlele: Actually no, will push with a `^`

sfc-gh-nlele: Actually no, will push a change for this

sfc-gh-nlele: Actually no, will push a change for this, but the package has been deprecated after this.

sfc-gh-afedorov (approved):

PR #598, opened 2022-03-11
Update manage.rst

sfc-gh-kmaurya:
Type error, duplicate "set set" -> "set"


sfc-gh-afedorov (approved):
lgtm, but do note that these docs are deprecated and will be deleted with the v1.10.0 release (i.e. they are not present in the v1.10.0 branch)

Issue #606, opened 2022-03-28
SNS handler specifies invalid region


ugmuka:
In AWS SNS handler, parameter `region_name` in boto3 client is read from environment variable `SNOWFLAKE_REGION`.
https://github.com/snowflakedb/SnowAlert/blob/8df0c9edde054463776fa58e88036ee2a783a41f/src/runners/handlers/sns.py#L31

However, AWS region is different from Snowalake region(ex. `us-west-2.aws`).
Boto3 region should be specified from different environment variable such as `AWS_DEFAULT_REGION`.

sfc-gh-afedorov:
good call, ty. if `AWS_SNS_HANDLER_REGION` is present, let's use that, otherwise fall back on `AWS_DEFAULT_REGION` and only if those two are missing fall back on `SNOWFLAKE_REGION`.

ugmuka:
Thank you for your review.
When do you plan to release v1.10.0?

sfc-gh-afedorov:
@sfc-gh-nlele may have a more concrete timeline towards release, but we auto-build `snowsec/snowalert:1.10-rc` off the release branch -- does that work all right for your use-case?

ugmuka:
No problem. I'll use `snowsec/snowalert:1.10-rc` image.
Best regards,

PR #607, opened 2022-03-29
add sns handler region(#606)

ugmuka:
fixed `boto3.client` region parameter in AWS SNS handler.

# related issue
#606


.whitesource
```diff
@@ -0,0 +1,15 @@
+{
+  "scanSettings": {
```
sfc-gh-afedorov, line 2: could we move this into a separate PR?

sfc-gh-afedorov (approved):

PR #660, opened 2024-03-04
GSE-3598

sfc-gh-kpatil:
Changing auth method with JAMF.
Moving to client credentials as JAMF is phasing out, basic auth.

this was tested by running the connector locally and verifying that the data is filled in `jamf_inventory_connection` table

### DIFF
--- a/src/connectors/jamf_inventory.py
+++ b/src/connectors/jamf_inventory.py
@@ -27,6 +27,11 @@ CONNECTION_OPTIONS = [
 HEADERS: dict = {}
 REQUEST_SPEED_PER_SECOND = 10
 
+CLIENT_ID_KEY = "client_id"
+CLIENT_SECRET_KEY = "client_secret"
+
+ACCESS_TOKEN_KEY = "access_token"
+
 
 async def fetch(session, url, wait=0) -> dict:
     if wait:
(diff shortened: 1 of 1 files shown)


src/connectors/jamf_inventory.py
```diff
@@ -75,7 +82,39 @@ async def main(table_name):
...
+        headers={"Content-Type": "application/x-www-form-urlencoded"},
+    )
+
+    json_response = response.json()
+
+    if ACCESS_TOKEN_KEY in json_response:
+        return json_response[ACCESS_TOKEN_KEY]
+
+    log.error(
+        f"{ACCESS_TOKEN_KEY} not found in response from api : {jamf_token_api_path}"
```
sfc-gh-afedorov, line 116: I'm not sure what `log.error` does in our case but it seems likely it's better to throw an exception so that it can be tracked in an error tracker like Sentry

src/connectors/jamf_inventory.py
```diff
@@ -75,7 +82,39 @@ async def main(table_name):
...
+
+# options is a dict containing cliendId and clientSecret
+def getAccessToken(credentials: str) -> str:
+
+    # credentials stored in format {"client_id":"some-id","client_secret":"some-secret"}
+    client_cred = json.loads(credentials)
+
+    jamf_token_api_path = "https://snowflake.jamfcloud.com/api/oauth/token"
+
+    data = {
```
sfc-gh-afedorov, line 98: there's no need to re-bundle these if you control the credentials -- just include the grant_type and pass them in with `json.loads(creds)` below

src/connectors/jamf_inventory.py
```diff
@@ -75,7 +82,39 @@ async def main(table_name):
...
+
+    if ACCESS_TOKEN_KEY in json_response:
+        return json_response[ACCESS_TOKEN_KEY]
+
+    log.error(
+        f"{ACCESS_TOKEN_KEY} not found in response from api : {jamf_token_api_path}"
+    )
+
+    # returning blank string  if access_token not found in response
+    return ""
```
sfc-gh-afedorov, line 120: it's better to raise an exception than return an empty string because there is nothing for the rest of the function to do with an empty string

src/connectors/jamf_inventory.py
```diff
@@ -75,7 +82,39 @@ async def main(table_name):
...
+    response = requests.post(
+        jamf_token_api_path,
+        data=urlencode(data),
+        headers={"Content-Type": "application/x-www-form-urlencoded"},
+    )
+
+    json_response = response.json()
+
+    if ACCESS_TOKEN_KEY in json_response:
+        return json_response[ACCESS_TOKEN_KEY]
```
sfc-gh-afedorov, line 113: use `access_token = response.json().get('access_token')` and then `assert access_token is not None, "no access token found in jamf oauth response"`, and whoever is looking at the error can find where this failure comes from to figure out the rest

src/connectors/jamf_inventory.py
```diff
@@ -75,7 +82,39 @@ async def main(table_name):
...
     return asyncio.get_event_loop().run_until_complete(main(f'data.{table_name}'))
 
+
+# options is a dict containing cliendId and clientSecret
+def getAccessToken(credentials: str) -> str:
+
+    # credentials stored in format {"client_id":"some-id","client_secret":"some-secret"}
+    client_cred = json.loads(credentials)
+
+    jamf_token_api_path = "https://snowflake.jamfcloud.com/api/oauth/token"
```
sfc-gh-afedorov, line 96: having this as a separate variable used once makes the code harder to read. inline it into the `requests.post` params

src/connectors/jamf_inventory.py
```diff
@@ -75,7 +82,39 @@ async def main(table_name):
...
+
+    data = {
+        'client_id': client_cred[CLIENT_ID_KEY],
+        'grant_type': 'client_credentials',
+        'client_secret': client_cred[CLIENT_SECRET_KEY],
+    }
+
+    response = requests.post(
+        jamf_token_api_path,
+        data=urlencode(data),
```
sfc-gh-afedorov, line 106: per [the docs](https://requests.readthedocs.io/en/latest/user/quickstart/#more-complicated-post-requests), you can pass a dict here and then you don't need to set headers or call urlencode.

sfc-gh-afedorov (approved):
this will probably work in the usual case but left some comments that will make the code more clear and succinct and work better in failure cases

sfc-gh-afedorov:
chatted just now and instead of `assert` we could also improve SnowAlert Data Connectors by creating a general `class AuthenticationError(RuntimeError)` and `raise AuthenticationError(...)` from the connector code

PR #671, opened 2024-05-20
Tag image with commit SHA after build, and push to repository

sfc-gh-gtan:
1. Added a post_push hook to add an extra tag using the commit SHA to the image after Dockerhub build finishes. This tag can help us with reverts in the future. Got this from: https://github.com/docker/hub-feedback/issues/341#issuecomment-248722131 and the docker hub documentation: https://docs.docker.com/docker-hub/builds/advanced/#push-to-multiple-repositories 

2. precommit was not working unless I set language_version to python3 instead of 3.7 locally. error below: 

precommit error


sfc-gh-gtan:
@sfc-gh-bkou @sfc-gh-afedorov: I'm not really sure how to test the post-build hook. 

I did not run the unit tests because my local env is not set up properly, and I wasn't able to run the tests

PR #692, opened 2026-04-13
fix(connectors): guard per-call/page failures so one error doesn't abort collection

sfc-gh-pkommini:
## Problem

An AWS region outage (me-south-1) causes `botocore.exceptions.EndpointConnectionError` during collection. This exception was not caught by `load_task_response`, which only handled `ClientError`, `DataNotFoundError`, and `ServerTimeoutError`. The uncaught exception propagates up and aborts the entire collection batch — all remaining accounts and regions are abandoned for that run.

## Fix

Add `EndpointConnectionError` to the except clause in `load_task_response` so connection-level failures from unreachable regions are recorded as error responses (the same way `ClientError` already is), instead of propagating and aborting the batch.

No changes to successful-path behavior.


src/connectors/aws_collect.py
```diff
@@ -1739,8 +1739,8 @@ async def load_pages():
             ):
                 yield x
 
-    # todo: double check whether these should be retried instead of recording errors
```
sfc-gh-afedorov, line 1742: keep this todo

src/connectors/aws_collect.py
```diff
@@ -1739,8 +1739,8 @@ async def load_pages():
...
-    except (ClientError, DataNotFoundError, ServerTimeoutError) as e:
+    except (ClientError, DataNotFoundError, BotoCoreError, ServerTimeoutError, ClientConnectionError) as e:
+        # BotoCoreError catches connection-level failures (e.g. EndpointConnectionError).
```
sfc-gh-afedorov: I think it might catch too much -- let's just continue on `EndpointConnectionError`

src/connectors/aws_collect.py
```diff
@@ -1775,7 +1775,7 @@ async def get_session(account_arn):
                 ),
             )
 
-        except ClientError as e:
+        except (ClientError, BotoCoreError, ClientConnectionError, ServerTimeoutError) as e:
```
sfc-gh-afedorov: I think the stack trace in the error shows that the session is retrieved successfully, probably because IAM is a global service, so we don't want to catch errors coming out of there, only from `load_task_response`

src/connectors/aws_collect.py
```diff
@@ -1802,29 +1802,51 @@ async def process_task(task, add_task) -> AsyncGenerator[Tuple[str, dict], None]
...
+                        async for response in load_task_response(client, task):
+                            if type(response) is DBEntry:
+                                if rn is not None:
+                                    response.entity['region'] = rn
+                                yield (task.method, response.entity)
+                            elif type(response) is CollectTask:
+                                add_task(response)
+                            else:
+                                log.info('log response', response)
+                except (BotoCoreError, ClientConnectionError, ServerTimeoutError) as e:
```
sfc-gh-afedorov: what raises these? the stack trace shows the exception through `load_task_response` and we will catch those with the change above

src/connectors/aws_collect.py
```diff
@@ -4,15 +4,15 @@
...
     ClientError,
     DataNotFoundError,
 )
 from aiobotocore.config import AioConfig
-from aiohttp.client_exceptions import ServerTimeoutError
+from aiohttp.client_exceptions import ClientConnectionError, ServerTimeoutError
 from collections import defaultdict, namedtuple
-import csv
 from datetime import datetime, timedelta
+import csv
```
sfc-gh-afedorov, line 14: alphabetical sorting is intentional here, let's keep it sorted that way

src/connectors/aws_collect.py
```diff
@@ -1802,29 +1802,51 @@ async def process_task(task, add_task) -> AsyncGenerator[Tuple[str, dict], None]
...
+                            },
+                        ),
+                    )
 
     else:
-        # record missing auditor role as empty account summary
+        # Record account-level error (missing role or unreachable region).
+        recorded_at = (
+            parse_date(e.response['ResponseMetadata']['HTTPHeaders']['date'])
+            if isinstance(e, ClientError)
```
sfc-gh-afedorov: hold off on `get_session` path modification for now -- it's a global service that isn't affected by the outage in Bahrain

src/connectors/aws_collect.py
```diff
@@ -1962,7 +1984,12 @@ def add_task(t):
 
             all_results: Any = defaultdict(list)
             for coro in asyncio.as_completed(coroutines):
-                result_lists = await coro
+                # One task failure shouldn't abort remaining collection tasks.
```
sfc-gh-afedorov: no, it should -- the contract between the connection driver and modules is that if the module cannot complete its ingestion, it should throw an exception. that's what drives the the ingestion dashboard where we review errors -- if we catch and silence errors here, it violates that contract and breaks that workflow.

sfc-gh-afedorov (changes requested):
left some feedback, and let's only fix AWSIC since that's the only thing that's breaking

src/connectors/aws_collect.py
sfc-gh-pkommini: Done, restored.

src/connectors/aws_collect.py
sfc-gh-pkommini: Sure, narrowed to just EndpointConnectionError

src/connectors/aws_collect.py
sfc-gh-pkommini: Makes sense, STS is global so get_session wouldn't hit this. Reverted, only catching in load_task_response.

src/connectors/aws_collect.py
sfc-gh-pkommini: EndpointConnectionError propagates through load_task_response which now catches it. Removed the redundant per-region guard.

src/connectors/aws_collect.py
sfc-gh-pkommini: Fixed, csv before datetime. Some sorta auto fix it seems like.

src/connectors/aws_collect.py
sfc-gh-pkommini: Reverted, keeping original except ClientError.

src/connectors/aws_collect.py
sfc-gh-pkommini: Understood, removed the catch-all in aioingest. Modules that can't complete should throw so the ingestion dashboard sees it.

sfc-gh-afedorov (approved):

